PR #1, opened 2015-10-22
Sign-Assembly function

nicolasvan:
Sign-Assembly is a function to sends a compiled assembly to a webserver so that the assembly can be digitally signed.

FYI: @nyctef, @samblackburn, @ChrisLambrou feel free to start using it in your build scripts once this is merged :smile:

### DIFF
--- a/Public/Sign-Assembly.ps1
+++ b/Public/Sign-Assembly.ps1
@@ -44,6 +44,9 @@ function Sign-Assembly {
   Add-ToHashTableIfNotNull $Headers -Key 'MoreInfoUrl' -Value $MoreInfoUrl
   Add-ToHashTableIfNotNull $Headers -Key 'ReCompressZip' -Value $ReCompressZip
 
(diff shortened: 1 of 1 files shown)


nyctef:
`$Server` should probably default to `rg-source01:8080` instead of being mandatory, I guess - I think everyone is going to give it the same value for now and this seems like one of the cases where being able to switch everyone at once over to a new url would be useful.

A bit of logging would be nice too - other than that it looks useful, thanks :)

nicolasvan:
This is a public repo though... do we want to leak this out ?

rmc47:
I'd agree with @nicolasvan that the less implementation-specific information we can leak, the better. I'm in the process of setting up SSL on the signing service, at which point it will also use the CNAME "signingservice.red-gate.com", albeit currently on port 4430 instead of 443, since SVN still uses 443 on that box :-(.

nyctef:
Ah, I didn't realise this was public

PR #4, opened 2015-10-31
Sign assembly improvements

ChrisLambrou:
Made Sign-Assembly more "PowerShelly", including the following:

- Renaming it to `Invoke-AssemblySigning`, to satisfy PowerShell naming conventions.
- Defined the help documentation correctly and expanded it to cover all parameters (well, the ones I understand).
- Renamed `AssemblyFilename` parameter to `AssemblyPath`, to better reflect that a full (or indeed relative) path is required.
- Removed the `FileType` parameter. The type is now automatically determined from the `AssemblyPath` parameter.
- Added `Parameter` attributes to the parameters, to provide some validation and make the usage a bit more flexible.
- Return the input `AssemblyPath` parameter, to enable cmdlet pipelining.


ChrisLambrou:
Questions:

1. What file-types are supported by the signing service? Under this change, only `.exe` and `.dll` are supported. Are there any others we should support, such as `.zip` or `.msi`?

2. What does the `ReCompressZip` actually do? If *we* don't know that, how can we expect users to know what to do with it?

nicolasvan:
I'm afraid I don't have any answer to these questions... :disappointed: 

_that's why I didn't dare touch any of these settings..._

ChrisLambrou:
Then perhaps we should restrict the cmdlet's api until someone comes along and asks about it. Alternatively, let's find out what the mean somehow.

nyctef:
Maybe we could just remove the parameter if [nobody's using it](https://github.com/search?utf8=%E2%9C%93&q=user%3Ared-gate+ReCompressZip&type=Code&ref=searchresults) - we can always add it back later

nyctef:
Found it in SVN: http://rg-source01/websvn/filedetails.php?repname=RedGate&path=%2FInternalProjects%2FBuildSystem%2FTrunk%2FSigningService%2FSigningService%2FSigning%2FVsixSigner.cs

nyctef:
This also appears to answer the first question - the valid `FileType`s are `Exe`, `Jar`, `Vsix` and `ClickOnce`

ChrisLambrou:
Hmm, I know I've seen `Dll` submitted, and I've seen it work. Am having a dig through the rest of the code to get a better idea of what it does.

ChrisLambrou:
Okay, I haven't seen `Dll` submitted after all. I'll make some more changes to this to support all the expected file types.

ChrisLambrou:
Okay, the FileType handling should now work for all the supported types, `.dll`, `.exe`, `.jar`, `.vsix` and `.application`. The key word here is *should*. We don't seem to have much in place for testing any of these scripts. I might have a look into Pester, a PowerShell testing framework. We ought to have tests in place for what will be a fairly critical part of our build infrastructure.

ChrisLambrou:
I've also documented the `ReCompressZip` parameter, but I'm actually inclined to remove it. It's currently only applicable to `.vsix` files (even though it could also apply to `.jar` files), and instructs the signing service to repack the zip file using maximum compression and specifically setting the deflate compression type of each zip entry. This doesn't feel like the signing service's responsibility and ought to be removed from it, IMO. Tools should just do one thing, and only one thing, really well. If people want their zip files optimised, they shouldn't rely on the signing service for this.

nyctef:
> I might have a look into Pester, a PowerShell testing framework. We ought to have tests in place for what will be a fairly critical part of our build infrastructure.

Sounds like it would be worth investigating, but I doubt we'd be able to any decent unit-testing here, since build scripts like these are fundamentally small integrations between other systems. 

> If people want their zip files optimised, they shouldn't rely on the signing service for this.

Totally agree with this

nicolasvan:
:+1: to all of it :smile:

ChrisLambrou:
I think Pester will let you mock out all parts of the PowerShell host environment. For example, `Sign-Assembly` takes a bunch of parameters and then calls `Invoke-WebRequest`. We can mock `Invoke-WebRequest` in order to write black-box regression tests for `Sign-Assembly`. Having those in place would make it easier for us to refactor and extend the cmdlet, with confidence that we're not breaking anything.

ChrisLambrou:
I've renamed `Sign-Assembly` to `Invoke-AssemblySigning`, but now that I understand it a bit more, it occurs to me that it deals with more than just assembly signing. Perhaps it should be called `Invoke-SigningService` instead?

nicolasvan:
Make sense, go for it

ChrisLambrou:
Okay, I've made the following changes, and apart from testing, I think I'm done:

1. Renamed the cmdlet to `Invoke-SigningService`, to indicate that it's more generic that just signing assemblies.
2. Renamed the input file parameter to `FilePath` to better reflect the more general usage. I've also defined aliases for this parameter (AssemblyPath, JarPath, VsixPath, ClickOnceApplicationPath), to make consuming scripts a bit more readable.
3. Removed the ReCompress parameter. This functionality doesn't belong in the signing service, so we don't want to encourage its use.
4. Added a couple of cmdlet help examples.

PR #5, opened 2015-11-02
Use new ConvertTo-ShellEscaped cmdlet.

ChrisLambrou:
Added a ConvertTo-ShellEscaped cmdlet and used it in Invoke-DotCoverForExecutable, where there was a TODO about needing improved shell escaping.


ChrisLambrou:
This requires some testing. I don't actually have any builds that use Invoke-DotCoverForExecutable.

ChrisLambrou:
I've added a whole bunch of tests for `ConvertTo-ShellEscaped`.

nyctef:
:+1: These look good - I think this is a much better example of where tests are useful compared to `New-TempDir`

nyctef:
Minor worry: Are there any tests which assert that multiple arguments don't get concatted into one long argument?

ChrisLambrou:
No. There's a single test that makes sure multiple arguments *are* concatenated into a single argument.

ChrisLambrou:
Yeah, the `New-TempDir` test was rather rubbish. It was just the easiest thing I could find to test! :smile:

nyctef:
> No. There's a single test that makes sure multiple arguments are concatenated into a single argument

I'm confused - if I pass `/arg1:foo /arg2:bar` into this function, does that mean it could get escaped into `"/arg1:foo /arg2:bar"` and end up as `args = new[] { "/arg1:foo /arg2:bar" }` in my `Main` function?

nicolasvan:
Can you just check that a call to Invoke-NUnitForAssembly with -EnableCodeCoverage still works ?

I remember spending soooo much time in powershell/dotcover character escaping hell....

we do need to make sure dotcover.exe still get the right parameters.... Easier said then done. :sweat:

ChrisLambrou:
The typical scenario is that you have an array of strings, say `[string[]] $Args`, you want those exact same strings to appear in your /NET application's `Main(string[] args)` function, and you're launching a process that accepts a single `arguments` string, then this cmdlet will handle that for you by escaping your string array, `$Args` to produce the single string value for `arguments`.

So yes, if you escape `'/arg1:foo /arg2:bar'` with this cmdlet, your Main function will receive `args = {"/arg1:foo /arg2:bar"}. If you pass `@('/arg1:foo', '/arg2:bar')` to this cmdlet, your Main function will receive `args = {"/arg1:foo", "/arg2:bar"}`.

I've used this in `Invoke-DotCoverForExecutable` as follows (loads of uninteresting bits removed):

```PowerShell
function Invoke-DotCoverForExecutable {
  [CmdletBinding()]
  param(
    ...
    [Parameter(Mandatory = $False)]
    [string[]] $TargetArguments,
    ...
  )

  $DotCoverArguments = @(...)
  ...
  # Here we have to manually conver the TargetArguments string array to a single escaped string.
  $EscapedTargetArguments = ConvertTo-ShellEscaped $TargetArguments
  $DotCoverArguments += "/TargetArguments=$EscapedTargetArguments"
  ...
  # Here PowerShell takes care of passing an array of strings to the executable.
  & $DotCoverPath $DotCoverArguments
  ...
}
```

Except of course, I've made a mistake here, which is why I wanted a code review, and someone who is using this cmdlet to actually test it for me! :smile: Fix to follow...

nyctef:
> So yes, if you escape `'/arg1:foo /arg2:bar'` with this cmdlet, your Main function will receive `args = {"/arg1:foo /arg2:bar"}`. If you pass `@('/arg1:foo', '/arg2:bar')` to this cmdlet, your Main function will receive `args = {"/arg1:foo", "/arg2:bar"}`.

Ah, I understand now. Cheers :)

ChrisLambrou:
@nicolasvan This cmdlet is primarily intended to help avoiding exactly the kind of PowerShell escaping hell that you've been though. As long as it actually works... heh heh!

nicolasvan:
As long as it works indeed....

If you get a compiled nunit dll you could easily test it using Invoke-NunitForAssembly and -EnableCodeCoverage...

Another think we'll have to test for is differences between Powershell v4 and v5... 

Very happy to see you take that on btw, good luck :+1: :pray:

ChrisLambrou:
Not sure how to automatically test it for PowerShell 4 and 5 other than to get TeamCity to run the build script twice. Need to make sure it doesn't double-publish the nuget package, though. Hmm...

nyctef:
Testing against 4 and 5 on the same machine might be tricky, since 5 appears to just replace 3 and 4
```
C:\scui buildscript/tests > $PSVersionTable.PSVersion.Major
5
C:\scui buildscript/tests > powershell -version 2 ' $PSVersionTable.PSVersion.ToString() '
2.0
C:\scui buildscript/tests > powershell -version 3 ' $PSVersionTable.PSVersion.ToString() '
5.0.10240.16384
C:\scui buildscript/tests > powershell -version 4 ' $PSVersionTable.PSVersion.ToString() '
5.0.10240.16384
```

nicolasvan:
Sounds like Microsoft means for v3,4,5 to be compatible.... Sadly there seems to be some variation... :disappointed:

nicolasvan:
I'm gonna build that package locally and use it to run the SQL compare engine unit tests. with and without coverage.... because dotcover...

Public/Invoke-DotCoverForExecutable.ps1
```diff
@@ -1,48 +1,65 @@
...
+  The arguments to pass to the target executable.
+
+.PARAMETER OutputFile
+  The output XML file containing the detailed coverage information.
+
+.PARAMETER DotCoverVersion
+  The version of dotCover nuget package to use.
+
+.PARAMETER DotCoverFilters
+  Coverage filters for dotCover, to indicate what should and should not be covered.
```
nicolasvan: you don't actually have to move the parameters to the top for the powershell help to kick in...
Powershell will find the comment on top of each parameter and display it in the generated help. 

Wouldn't keeping these comments along with the parameters make it easier to keep the doc up to date?

nicolasvan:
Well it works on my win 10 machine (powershell v5)

I can execute Invoke-NUnitForAssembly with -EnableCodeCoverage and some excluded categories and all parameters are passed correctly to dotcover.exe (which then passes them on correctly to nunit...)

I'm impressed. :clap:

RedGate.Build.nuspec
```diff
@@ -15,6 +15,7 @@
   <files>
     <file src="Private\*.ps1" target="tools\Private" />
     <file src="Private\*.psm1" target="tools\Private" />
+    <file src="Private\nuget.exe" target="tools\Private" />
```
nicolasvan: we need that guy :smile:

_Yeah a nuget package containing nuget.exe :expressionless:  _

Public/Invoke-DotCoverForExecutable.ps1
ChrisLambrou: Sure. Sounds like a good idea. I was just following the documentation pattern from some MS PowerShell guide, but if it works when you keep the comments closer to the cmdlet parameters, that's definitely nicer.

RedGate.Build.nuspec
ChrisLambrou: I know. I kept it there because that's what the previous NuSpec definition did. I'd be delighted to get rid of it, but that wasn't the focus of what I was trying to achieve on the Pester branch.

Can't we get rid of it? Pretty please? With sugar on top?

ChrisLambrou: I'm confused. I didn't mean the Pester branch. I meant this one.

nicolasvan: if we remove it, we can't easily call nuget.exe to restore Nunit.Runners and dotcover from within RedGate.Build

https://github.com/red-gate/RedGate.Build/blob/master/Public/Install-NUnitPackage.ps1
https://github.com/red-gate/RedGate.Build/blob/master/Public/Install-Package.ps1

Install-NUnitPackage ends up being called by Invoke-NUnitForAssembly...

I'm open to suggestion :smile:

ChrisLambrou: What's especially annoying is that most consumers of `RedGate.Build` will already have a `nuget.exe` file available, since they must have used it to obtain this module in the first place.

We could write a `Get-NuGetPath` cmdlet that returns the path of an appropriate version of `nuget.exe`, for this module's own use. Internally it could use some strategy to obtain it, perhaps the following:

1. If `.\Private\nuget.exe` exists, and is an acceptable version, return it.
2. Search up the path to the `.build` folder and see if we can find a copy of `nuget.exe` (the one that was probably used to obtain this module in the first place). If it's an acceptable version, copy it to `.\Private\nuget.exe`, then go back to step 1.
3. Download `nuget.exe` from nuget.org to `.\Private\nuget.exe`, then go back to step 1.

ChrisLambrou: Oh, hang on a minute. I've misunderstood this. Aside from the slightly icky feel to it, there's no problem in principle with including a binary dependency in a NuGet package. We could add a package dependency, but given the way we're recommending that `RedGate.Build` is installed, I think it's better to avoid those and opt for bundling instead.

nyctef: > Internally it could use some strategy to obtain it, perhaps the following

There's also a fourth alternative  - read the teamcity config for `teamcity.tool.NuGet.CommandLine.2.8.3.nupkg` to find the `nuget.exe` that teamcity has already installed on the build agent. For Compare I ended up just checking it in since it seemed too much hassle otherwise though :disappointed:

ChrisLambrou: That would make local builds different from TC builds, which is something we're frowning upon, isn't it?

I don't see much problem with including `nuget.exe` in the RedGate.Build nuget package. It's not like it's being checked into a git repo anywhere (indeed it's not in *this* repo), and a package restore from the local cache should be suitably fast given that we're hoping for lots of builds to be using this package.

nicolasvan: It feels to me that adding some code to try to get hold of nuget.exe (even if we know that there must be nuget.exe somewhere to be able to use RedGate.Build.nupkg in the first place) is not worth the hassle, and might even make things more complicated....

and I agree that using `teamcity.tool.NuGet.CommandLine.2.8.3.nupkg` means a different build on our local machine and ou CI server... which I DO frown upon :frowning: 

I would suggest to keep it like this until this starts becoming a problem :smile: 

*It's still good to be aware that RedGate.Build uses its own version of nuget.exe while the build script using RedGate.Build may be using another version of nuget.exe*

nyctef: Well, the assumption was that all versions of `nuget.exe` called `2.8.3` would be equivalent, so you'd just need to get it from *somewhere* and could treat `teamcity.tool.NuGet.CommandLine.2.8.3.nupkg` as a cache for it. But yeah, the extra complexity and teamcity/local differences that it introduced made me think it wasn't worth it in this case

nyctef: Welp, I've just run into a problem - the latest `RedGate.Build` nuget packages are broken without this `nuget.exe` (since `Private/_Init.ps1` tries to resolve the path, but it ain't there).

Could we merge this soon, please? :D

PR #6, opened 2015-11-02
Adds Pester tests

ChrisLambrou:
Introduced Pester testing.


nyctef:
Might be worth looking into `-EnableExit` and `-OutputFile` from Pester (and `TeamCity-ImportNUnitReport` from this repo) so we can get the results into the teamcity build output

ChrisLambrou:
I tried adding a build feature to automatically monitor the TestResults.xml file, but it's a bit rubbish. If we don't clean up the files from the previous checkout, then it monitors the results from a previous build. Even worse, whilst it's doing this, it sometimes locks the results files, and the build fails when trying to clean it. I'll look into using `TeamCity-ImportNUnitReport`.

nyctef:
> If we don't clean up the files from the previous checkout, then it monitors the results from a previous build

Could this be because [you've checked the results file in?](https://github.com/red-gate/RedGate.Build/pull/6/files#diff-b39fb95e25db35acd744a30bc371e079) :P

ChrisLambrou:
> Could this be because you've checked the results file in? :P

Maybe.... :P

ChrisLambrou:
But even if the file is there, the build script tries to delete it, and the build barfs because TeamCity is monitoring the file. Doesn't seem to matter how the file got there (previous build or accidentally checked in).

ChrisLambrou:
> Might be worth looking into -EnableExit and -OutputFile from Pester (and TeamCity-ImportNUnitReport from this repo) so we can get the results into the teamcity build output

-EnableExit turned out to be a bit of a disaster. It causes the current process to terminate with an exit code equivalent to the number of failing tests. When the current process is the PowerShell host, that's hugely annoying! :disappointed:

nyctef:
> It causes the current process to terminate with an exit code equivalent to the number of failing tests. When the current process is the PowerShell host, that's hugely annoying!

Ah, I see - I assumed it's what we wanted (since it's the same behaviour that `nunit-console` has) but I forgot we would actually be running in the build script's powershell process instead of a subprocess :D

ChrisLambrou:
I think this is probably done. The aim wasn't to actually write lots of Pester tests, but to simply get the infrastructure in place.

ChrisLambrou:
The NuGet package produced seems okay. Going to merge this. I plan to change the build so that it's the NuGet package that's tested, not the original files, but I'll do that on another branch.

PR #8, opened 2015-11-06
New-NuGetPackageVersion cmdlet

ChrisLambrou:
Initial import of the Get-NuGetPackageId cmdlet. It currently does a whole bunch of stuff, some of which ought to be unpicked, and some of which should stay here.

Also, there are no tests yet, but I'll write as we decide how this thing ought to actually work.


nyctef:
Just pasting [initial comments from slack](https://redgate.slack.com/archives/build-scripts/p1446813378000011):

> 
Mark Jordan [12:36 PM] 
chrislambrou: Good idea, although I think that function needs to be split up /a lot/ before it's reusable since it knows way too much at the moment (that the BRANCH_NAME and IS_DEFAULT_BRANCH env vars are set appropriately; how to shorten a branch name into a valid pre-release suffix; that the global $Version variable is set to a valid System.Version; that we only want the first three numbers from $Version to be in the package version and that we want to put $Version.Revision into the pre-release string)
>
>Chris Lambrou [12:51 PM] 
Well the whole point is to take on that burden, so callers of the cmdlet don't need to worry about it. I'd probably want parameters for the version number, the branch name, and whether or not the current branch is the default branch (probably with some default values taken from environment variables). As for using using semantic version numbers and shortening the name, the former is good practice and the latter is absolutely necessary.
>
>I'll create a branch for this, so we can discuss it further and make changes.
>
>Mark Jordan [12:56 PM] 
>The problem is not that any of them are really bad ideas (although I've been thinking about alternatives to environment variables, and our nuget packages simply won't work with three-part version numbers at the moment) it's the fact that if I have to disagree with any one of those things, I have to throw away the whole function and write my own, so it becomes much less useful as a thing to share. Splitting up responsibilities for a function like that makes it much more likely that I can actually use it.

ChrisLambrou:
Well, first thing's first. Describe what this cmdlet does support, and consider what it doesn't currently support.

**What it does:**
Based on the assumption that build numbers will be of the form x.x.x.y, where y is a numerically increasing build number, it does the following.

1. Given a 4 digit version number, say 1.2.3.4, it will return '1.2.3.4' on the default master branch.
2. For branches other than master, say 'foobar', it will return 1.2.3-foobar4.
3. It sanitizes the branch name for use as a pre-release suffix
   1. invalid characters (e.g. underscores) are simply excluded.
   2. if the name is still too long, it does its best to shorten it.
4. It automatically determines the branch name, whether or not the branch is the default branch, and whether or not it's an automated build.
   1. The branch name is taken from the `BRANCH_NAME` environment variable, on the assumption that TeamCity will establish this property on build agents.
   2. Whether or not the branch is the default branch is taken from the `IS_DEFAULT_BRANCH` enviornment variable, again on the assumption that TeamCity will establish this property.
   3. The combined presence of the two aforementioned environment variables is used to determine if the build is automated (i.e. on a build agent) or not (i.e. a local developer build). This in turn overrides the pre-release suffix, setting it to '-dev', on the assumption that only TeamCity builds should be published, and that local builds should never be published.

**What it doesn't do**
It doesn't support version numbers with fewer than 4 digits. Version numbers are expected to be of the form `major.minor.patch.build' (this is consistent with semantic versioning and something we should be striving for).

ChrisLambrou:
Things I definitely thing we should unpick from this cmdlet:

1. Automatically determining the branch name.
2. Automatically determining if the branch is the default branch.
3. Special treatment for local builds.

Items 1 and 2 should be supported via cmdlet parameters. I'm undecided about whether they should default to environment variables set by TeamCity. There's already a precedent set for this with the Signing Service URL in the `Invoke-SigningService` cmdlet, but I can't help thinking we should think more deliberately about conventions for running in TeamCity.

Item 3 should simply be removed.

ChrisLambrou:
What I don't want to remove from this cmdlet is the automatic derivation of the pre-release suffix from the branch name. Using the branch name, and whether it's the default branch, to establish a pre-release suffix, and dealing with the pre-release string character and length restrictions, is essentially the whole point of this cmdlet.

PR #10, opened 2015-12-04
Add a default signing url

adrianbanks:
This sets the signing url to a default value if one isn't specified.


nyctef:
@adrianbanks: See https://github.com/red-gate/RedGate.Build/pull/1 for why we didn't do this originally

nicolasvan:
Thanks for the PR @adrianbanks :smile: 
The thing is that right now, the repo is public.... and we didn't feel like leaking this Url in...

nicolasvan:
So of course, we could change our minds and make this repo private ?
But keeping it public might help us keep it clean ? And showcase (_a little bit_) how we do things here?

adrianbanks:
Ah, ok. Didn't notice it was public. I'll try a different approach then.

Perhaps it is worth adding a `contributing.md` with some guidelines on what should and shouldn't be in here.

PR #21, opened 2016-03-03
New Update-AssemblyVersion cmdlet.

ChrisLambrou:
Cmdlet that uses regular expression magic to update the AssemblyVersion, AssemblyFileVersion and AssemblyInformationalVersion attributes in a source file, typically AssemblyInfo.cs.

### DIFF
--- a/Public/ConvertTo-ShellEscaped.ps1
+++ b/Public/ConvertTo-ShellEscaped.ps1
@@ -1,4 +1,3 @@
-#requires -Version 2
 <#
 .SYNOPSIS
   Shell-escapes one or more strings for use in a command-line when starting a process.
@@ -17,6 +16,7 @@
 
   -path "C:\Program Files\\" -db "Data Source=local;Application Name=\"My app\""
 #>
+#requires -Version 2
 function ConvertTo-ShellEscaped
 {
     [CmdletBinding()]
--- a/Public/ConvertTo-ShellEscaped.ps1
+++ b/Public/ConvertTo-ShellEscaped.ps1
@@ -1,30 +1,21 @@
 #requires -Version 2
-
 <#
-        .SYNOPSIS
-        Shell-escapes one or more strings for use in a command-line when starting a process.
+.SYNOPSIS
+  Shell-escapes one or more strings for use in a command-line when starting a process.
+.DESCRIPTION
+  Given one or more argument strings, for each argument, this function will escape double-quote characters and some backslash characters, and surround the string with double-quotes if necessary. The results are then concatenated, using a single ' ' separator character.
 
-        .DESCRIPTION
-        Given one or more argument strings, for each argument, this function will escape double-quote characters and some backslash characters, and surround the string with double-quotes if necessary. The results are then concatenated, using a single ' ' separator character.
+  See http://msdn.microsoft.com/en-us/library/a1y7w461.aspx for details on the necessary escaping.
+.PARAMETER Arguments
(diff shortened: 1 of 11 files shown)


nyctef:
Minor complaint: looks like this'll write out with utf8+BOM encoding by default, which is [kind of icky](http://utf8everywhere.org/#faq.boms). I don't know if we have (or should have) any general rules on file encoding, though.

nyctef:
Removing the two 

```
        Author: Chris Lambrou
        Copyright 2016, Red Gate Software Limited
```
lines makes `Get-Help Update-AssemblyVersion` work and the tests pass - do we need those? (maybe it could go into `.DESCRIPTION` or `.NOTES`)

ChrisLambrou:
I just copied them from another file. Should get rid of them everywhere if they break Get-Help.

ChrisLambrou:
Regarding the bom, we could make a special case for utf8 and use File.WriteAllText with and appropriately crafted Encoding instance. Maybe another private helper cmdlet as an alternative to Out-File?

nicolasvan:
For the BOM thing, I would keep using powershell `-Encoding UTF8` and trust powershell to do the right thing. BOM or not BOM, as long as this doesn't break our builds we are probably fine....

Maybe one day Microsoft will remove the BOM ?

@nyctef the link you give also says we should use `\n` even on windows, which we probably won't follow either :stuck_out_tongue:

ChrisLambrou:
What link?

nicolasvan:
http://utf8everywhere.org/#faq.boms 
http://utf8everywhere.org/#faq.crlf

nyctef:
> @nyctef the link you give also says we should use `\n` even on windows, which we probably won't follow either :stuck_out_tongue: 

windows newlines (ie CRLF) are also dumb and terrible, but that's another holy war for another time :stuck_out_tongue:

ChrisLambrou:
Meh to boms and line endings. I've gone through and tidied up all the help docs a bit (not just my new cmdlet).

ChrisLambrou:
Ah, that link! Sorry, was up late following the SF saga and am a bit bleary eyed this morning. :|

ChrisLambrou:
Okay, the documentation tests now pass. Turns out that the `Get-Help` cmdlet is truly rubbish. It basically insists that the special code comments are the first thing in the file, so it was failing on the two cmdlet that had `#requires -Version X` directives automatically inserted at the top of the file by PowerShell ISE.

Anyway, still not ready to merge. Although I tried it out last night on another project and it basically appears to work okay, I haven't written any tests for this, yet, which I intend to do tomorrow sometime.

nyctef:
It's only doing what the docs say it does :stuck_out_tongue: 

https://technet.microsoft.com/en-us/library/hh847834.aspx
> Comment-based help for a script can appear in one of the following two
    locations in the script.
>    -- At the beginning of the script file. Script help can be preceded in the
       script only by comments and blank lines.
>    -- [...]

ChrisLambrou:
That `#requires -Version` directive *is* a comment! I stand by my criticism of `Get-Help`. I found dozens of other bugs in it when I wrote the `XmlDoc2CmdletDoc` tool, too.

nyctef:
Nah, it's just [a statement that sneakily looks like a comment](https://technet.microsoft.com/en-us/library/hh847765.aspx) :wink:

ChrisLambrou:
I've added some tests and corresponding bug fixes. :shipit:?

nicolasvan:
:+1:

PR #22, opened 2016-03-09
Support 3 part version numbers with feature branches, also fixup inco…

richardjm:
…rrect example, add new example, add tests for three part


nicolasvan:
:+1:

Issue #23, opened 2016-03-09
Update-AssemblyVersion adds new-lines to the end of files


ChrisLambrou:
It seems that `Update-AssemblyVersion` is adding sneaky extra lines of white-space to the end of the updated source file. This was being masked by a test that was accounting for behaviour in an earlier incarnation of the implementation.

Originally reported by @richardjm

nyctef:
Is this http://stackoverflow.com/questions/35279848/prevent-trailing-newline-in-powershell-out-file-command?

richardjm:
Write a PR and remove Trim() to find out :)

ChrisLambrou:
The more I work with PowerShell, the more I get pissed off with `Get-Content` and 'Out-File`. I think I'll resort to using `File.ReadAllText` and `File.WriteAllText` instead. I can also fix a BOM issue, too.

richardjm:
Resolved with the PR

ChrisLambrou:
Glad it helps. Please keep the branch open, though. I want to write some more tests. :)

richardjm:
Can you create a new branch from experience github doesn't like to take two PRs from the same branch - especially if it was recreated. Can't see a way to un-delete it from github.

ChrisLambrou:
Okay, will do.

PR #24, opened 2016-03-09
Fix unwanted whitespace emitted by Update-AssemblyVersion

ChrisLambrou:
This addresses #23

We now use System.IO.File static methods ReadAllText and WriteAllText to
load and save the source files, to avoid problems with new-line
characters and unwanted extra whitespace.

The type of the `Encoding` partameter has been changed from `string` to `System.Text.Encoding`. The default is still to use UTF8 sans BOM.

The `$SourceFilePath` parameter now also supports wildcards.


Private/Get-Encoding.ps1
```diff
@@ -0,0 +1,45 @@
+<#
+.SYNOPSIS
+  Updates the AssemblyVersion, AssemblyFileVersion and AssemblyInformationalVersion attributes in a source file.
```
richardjm: Synopisis and description here are incorrect.

Public/Update-AssemblyVersion.ps1
```diff
@@ -45,20 +55,30 @@ function Update-AssemblyVersion
         [string] $Encoding = 'UTF8'
     )
 
+    # Resolve SourceFilePath to an absolute path, and make sure it exists.
+    $SourceFilePath = Resolve-Path $SourceFilePath
```
richardjm: Resolve-Path allows wildcards - what would the effect be on passing wildcards into this routine? I already work around this behaviour in a build script - https://github.com/red-gate/RedGate.Logging/blob/master/RedGate.Logging.Interfaces/.build/build.ps1#L68

Public/Update-AssemblyVersion.ps1
```diff
@@ -12,7 +12,17 @@
  .PARAMETER InformationalVersion
   The optional Assembly Informational Version to be injected into the source file. If unspecified, defaults to the value of the Version parameter.
 .PARAMETER Encoding
-  The optional encoding of the source file. If unspecified, defaults to 'UTF8'. Valid values include 'Unicode', 'Byte', 'BigEndianUnicode', 'UTF8', 'UTF7', 'UTF32', 'Ascii', 'Default', 'Oem' and 'BigEndianUTF32'.
```
richardjm: Can we make a breaking change so that this parameter is System.Text.Encoding, rather than have the enum map in Get-Encoding.ps1 at all.

richardjm:
Also - I know it's not new but you've modified behaviour - there are no encoding tests at all.

Public/Update-AssemblyVersion.ps1
nyctef: The problem here (and the main reason for `Get-Encoding` to exist as I understand it) is that creating a utf8-without-BOM encoding is actually rather annoying, so it would be nice to take that pain away from consumers of `Update-AssemblyVersion`

Public/Update-AssemblyVersion.ps1
ChrisLambrou: Yeah, will rewrite that with `Join-Path' and 'Get-Location'.

nicolasvan: I would keep Resolve-Path as it is quite clear what it does.
If someone passes in wildcards, and it breaks, then I'm pretty they will stop passing wildcards in on their own ?

Public/Update-AssemblyVersion.ps1
ChrisLambrou: Yes, that's the problem, sadly. I didn't really want to break the API, but sod it - no one else is really using this yet. I'll change the type of the Encoding parameter to `System.Text.Encoding`, and consider making Get-Encoding public. Encoding should still be optional, and default to UTF without BOM, though.

Private/Get-Encoding.ps1
ChrisLambrou: If the parameter is of type `System.Text.Encoding`, optional, and defaults to UTF8 without BOM, then there's probably no need for the `Get-Encoding` cmdlet. Will probably get rid of it.

Public/Update-AssemblyVersion.ps1
ChrisLambrou: Or I could make it work with wildcards?

nicolasvan:
```
@(Resolve-Path $SourceFilePath) | ForEach {
   # do your thing with $_ ?
}
```
yep that could work :)

richardjm: Please - that would be nice - can I specify a wildcard which is something like **/AssemblyInfo.cs in powershell?

nicolasvan: `**` are not supported in Powershell :disappointed: 
```
Get-ChildItem 'myfolder' -Filter AssemblyInfo.cs -Recurse | Select FullName | Update-AssemblyInfo -Version 'blah'
```
Is probably the shortest alternative ? (not 100% sure whether we do need the `Select Fullname`)

richardjm: Pretty much bang on what I've done already in https://github.com/red-gate/RedGate.Logging/blob/master/RedGate.Logging.Interfaces/.build/build.ps1#L68

```
task SetAssemblyVersion {
    Get-ChildItem "$RootDir" "AssemblyInfo.cs" -Recurse | foreach {
        Update-AssemblyVersion `
             $_.FullName `
            -Version $AssemblyVersion `
            -FileVersion $AssemblyFileVersion `
            -InformationalVersion $NuspecVersion `
            -Verbose
            } | Out-Null
}
```

nicolasvan:
This small PR is becoming a big discussion, so hey I'll add my 2 cents :stuck_out_tongue: 

I would hardcode the encoding to `UTF8` 
We just need the encoding so that we don't break the `copyright` symbol. That's all.
I would not let the users override it at all, because there is no need.

BOM or not BOM, we really don't need to care as long as the AssemblyInfo.cs still gets compiled properly. So unless csc.exe does not support BOM, I would keep using the easier to read `-Encoding UTF8` or (`[System.Text.Encoding]::UTF8`).

and move on ! :smile: :stuck_out_tongue:

Public/Update-AssemblyVersion.ps1
ChrisLambrou: I don't think `select FullName` is required. The child items are `FileInfo` instances. PowerShell converts them to `string` to satisfy the `SourceFilePath` parameter, which is equivalent to the full name? I'll just have to try it. I don't have any tests for multiple files, yet.

nyctef:
> BOM or not BOM, we really don't need to care as long as the AssemblyInfo.cs still gets compiled properly.

Actually, I really do care, because of shit like this everywhere which git keeps highlighting as errors:

image

Public/Update-AssemblyVersion.ps1
nicolasvan: Agreed it should not be needed, but I suspect it might actually convert FileInfo to just Name (and thus lose the directory....) definitely worth trying out :)

nicolasvan:
I insist we don't care about BOM. we're human beings and we can easily overlook the ugly `<U+FEFF>`. We're not really editing AssemblyInfo.cs ever, so we're not suffering from it.

Maybe use a different git diff viewer ?

ChrisLambrou:
The `Encoding` parameter is now optional, of type `System.Text.Encoding`, and does default to UTF8 sans BOM. The implementation for dealing with this isn't particularly complicated or hard to read, so I hope it keeps everyone happy.

nicolasvan:
fine

PR #27, opened 2016-04-29
Fix Issue 26 - Install-Package copes with version differences

ChrisLambrou:
`Install-Package` now copes with the possibility that NuGet might not install the precise version that was requested. e.g. if version 1.2.3 is asked for, NuGet may actually install version 1.2.3.4.

The solution is to allow NuGet to do it's thing, then scan the packages folder to see what was actually installed, and return the correct folder, rather than guessing the folder based solely on the package id and requested version.

### DIFF
--- a/Public/Install-Package.ps1
+++ b/Public/Install-Package.ps1
@@ -1,29 +1,52 @@
 <#
-.SYNOPSIS
-  Install a Nuget Package to the RedGate.Build\packages\ folder
-.DESCRIPTION
-  Install a Nuget Package to the RedGate.Build\packages folder
-  and return the full path of the folder where the package was extracted to.
+        .SYNOPSIS
+        Install a Nuget Package to the RedGate.Build\packages\ folder.
+        .DESCRIPTION
+        Install a Nuget Package to the RedGate.Build\packages folder
+        and return a DirectoryInfo object for the full path of the folder
+        where the package was extracted to.
+        .PARAMETER Name
+        The name/id of the nuget package to install.
+        .PARAMETER Version
+        The version of the nuget package to install.
+        .OUTPUTS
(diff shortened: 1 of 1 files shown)


ChrisLambrou:
I tried scraping the console output of NuGet, but it was messy and unreliable. Scanning the packages folder for a package that matches the requested package name and version is much more reliable.

nicolasvan:
:+1:

ChrisLambrou:
Cheers, ears! :smile:

richardjm:
Looks like returning `DirectoryInfo` instead of the full path to the directory is breaking New-ZipArchive... 
@ChrisLambrou any specific reason we need to return a directoryinfo ?

This is the line that falls apart https://github.com/red-gate/RedGate.Build/blob/master/Private/Zip.ps1#L7

```
"$7ZipFolder\tools\7za.exe" | Resolve-Path
```
$7ZipFolder is not the fullpath anymore....

Looking at this with Nico at the moment, @ChrisLambrou can we change this to return the full path again? or will it break the world on your side ?

richardjm:
I've raised a PR to revert this behaviour - https://github.com/red-gate/RedGate.Build/pull/30

ChrisLambrou:
`Resolve-Path` doesn't return a string, it returns a `PathInfo`. I thought I was just following that pattern, but got confused between `PathInfo`, which is what I though the new version was returning, and `DirectoryInfo`, which is what it actually returns. Am more then happy to return a `PathInfo` instead. I'd be even happier if we returned an actual string, as described in the original doc comment.

ChrisLambrou:
Okay, I commented on #30 in response to being mentioned here, but someone already merged it. My issue still stands - `Install-Package` returns a `PathInfo` but its doc comment claims it returns a string. This is misleading and should be resolved.

PR #28, opened 2016-04-29
Added Install-NodeJsPackage cmdlet.

ChrisLambrou:
This provides a new cmdlet, `Install-NodeJsPackage`, which will install nodejs using the `RedGate.ThirdParty.NodeJs` package, returning an object with the following string properties:

``` PowerShell
@{
    NodeJsVersion = '6.0.0'; # The installed version of nodejs.
    NodeJsDir = 'C:\Tools\NodeJs\6.0.0'; # The folder nodejs was unpacked to.
    NodeExePath = 'C:\Tools\NodeJs\6.0.0\node.exe'; # The full path of node.exe.
    NpmCmdPath = 'C:\Tools\NodeJs\6.0.0\npm.cmd'; # The full path of npm.cmd.
}
```

You could use it as follows:

``` PowerShell
$NodeExePath = (Install-NodeJsPackage6.0.0).NodeExePath
```

Note that the `RedGate.ThirdParty.NodeJs` is not yet available in the public nuget.org package source. Hopefully that will happen after some internal testing. Meanwhile, current users of this cmdlet will have to make sure that the internal Redgate package source is specified in their local `NuGet.config`.

### DIFF
--- a/Private/_Init.ps1
+++ b/Private/_Init.ps1
@@ -12,6 +12,7 @@ $NugetExe = Resolve-Path "$PSScriptRoot\nuget.exe"
 $DefaultNUnitVersion = '2.6.4'
 $DefaultDotCoverVersion = '3.2.0'
 $DefaultSmartAssemblyVersion = '6.8.0.248'
+$DefaultNodeJsPackageVersion = '6.0.0'
 
 
 if ($Host.Name -eq "Default Host") {
--- a/Public/Install-NodeJsPackage.ps1
+++ b/Public/Install-NodeJsPackage.ps1
@@ -1,0 +1,50 @@
+<#
+        .SYNOPSIS
+        Installs a working copy of nodejs and npm.
+        .DESCRIPTION
+        Installs the RedGate.ThirdParty.NodeJs package to RedGate.Build\packages,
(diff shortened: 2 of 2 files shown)


deneboulton:
Hey Chris,

I tried using this on the build server:
http://buildserver/viewLog.html?buildId=7328589&buildTypeId=Shared_RedGateClient_2ServicePluginsShims&tab=buildLog&state=80#_state=80&focus=70

Which seems to point at this:
https://github.com/red-gate/RedGate.ThirdParty.NodeJs/blob/9e5ebe24a687e1dd3763adad4d087a17758b8296/install.ps1#L18

Perhaps is the static method ‘new’ (::new) a Powershell v5 feature? I believe our build machines are on v4.

deneboulton:
Oh, and the cmdlet seems to be Install-NodeJsPackage, not Install-NodeJs like in the example ;)

ChrisLambrou:
Cheers Dene. The `::new` syntax appears to be specific to PowerShell 5, and I have a fix for that. I also spotted the name mismatch, too, but I can't remember offhand where I've fixed it. Probably forgot to push my changes before I left for the weekend, but will try to find time over the weekend to finish off the fixes.

ChrisLambrou:
This is completely stale. Killing it.

PR #30, opened 2016-05-05
Install-Package fix breaking change to return fullpath as string not …

richardjm:
…directory info, zip allow to function without basepath argument


ChrisLambrou:
This has already been merged, but I still have comments to make on it. Either `Install-Package` should be modified to return a string, or it's doc comment should be modified to indicate that it returns a `PathInfo`. I kinda prefer the former, but we ought to do one of these.

PR #31, opened 2016-05-05
Install-Package to actually return a string. As per its own documentation

nicolasvan:
As per https://github.com/red-gate/RedGate.Build/pull/30#issuecomment-217125149

+ a couple of tests just in case...


ChrisLambrou:
All looks good to me. :)

PR #33, opened 2016-06-22
testing testing

nicolasvan:
testing notifications


nicolasvan:
Hmm teamcity user DOES need write access to the repo.
As per https://developer.github.com/v3/repos/statuses/#create-a-status

> Users with push access can create commit statuses for a given ref

PR #35, opened 2016-07-06
The console exe moved packages for 3.0.0.0+

GrahamTheCoder:
See http://buildserver.red-gate.com/viewLog.html?buildId=7483806&tab=buildResultsDiv&buildTypeId=SQLClone_Build

Anyone else please feel free to pick this up. It's still failing with 

```
Invalid argument: /nologo
Invalid argument: /nodots
Invalid argument: /noshadow
The value '/out:C:\Repos\GitHub\sqlclone\RedGate.SqlClone.DataAccess.IntegrationTests\bin\Release\RedGate.SqlClone.DataAccess.IntegrationTests.dll.TestResult.TestOutput.txt' is not valid for option '--labels'.
```

I don't have time to fix it just now, but I was testing it out on SqlClone https://github.com/red-gate/SqlClone/compare/add-coverage?expand=1

### DIFF
--- a/Private/NUnit.ps1
+++ b/Private/NUnit.ps1
@@ -45,11 +45,19 @@ function Get-NUnitConsoleExePath {
         #By default, use nunit-console.exe
         [switch] $x86
     )
-
-    $nunitExec = 'nunit-console.exe'
+	
+	$nunitBase = 'nunit'
+	if (!$NUnitVersion.StartsWith("2.")) {
+		# Version 3 puts the major version in front, let's assume that's a trend
+		$nunitBase += $NUnitVersion.Split(".")[0]
+	}
+	
+	$nunitConsole = '-console'
     if($x86.IsPresent) {
-        $nunitExec = 'nunit-console-x86.exe'
-    }
+        $nunitConsole += '-x86'
+    }	
+	
+    $nunitExec = $nunitBase + $nunitConsole + '.exe'
 
     Write-Verbose "Using NUnit version $NUnitVersion"
     $NUnitFolder = Install-NUnitPackage $NUnitVersion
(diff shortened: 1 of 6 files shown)


nyctef:
It looks like everything changes for NUnit 3 - `Build-NUnitCommandLineArguments` changes (because of the `Invalid argument /nologo ...` stuff), `Get-NUnitConsoleExePath` changes (because it now has a version number) and `Install-NUnitPackage` changes (because it's a different package)

I think it would probably end up just being simpler to have `NUnit3` versions of all the cmdlets instead of adding `if`s everywhere.

GrahamTheCoder:
Yep this branch probably needs killing, `if`s everywhere is not what we want, I was just following the trail of error messages.

nyctef:
Out of curiosity, how are you finding NUnit 3? Is it significantly better than NUnit 2 that we should look at switching?

GrahamTheCoder:
I'm not aware of anything we currently gain from using it. Our default position is to use the latest major version of everything. In that vein we'll possibly switch to .NET Core which I think would require [NUnit 3](http://www.alteridem.net/2016/06/18/nunit-3-testing-net-core-rc2/)

If it's lots of faff I can live with downgrading it for now.

nicolasvan:
Damn, just seeing the comment now after I touched everything...
It is pretty ugly, but there are some tests now....

what do you guys think?

nicolasvan:
hmm the way we can select tests to execute is also completely different. (there does not seem to be any support for exluding/including categories)
`--exclude`, `--include` are not valid parameters. Instead we should probably use `--where EXPRESSION`

I guess we might need a brand new `Invoke-NUnit3ForAssembly` function and clean everything up?

GrahamTheCoder:
+ 10 cool points for the tests. Yeah sounds like it's worth a new top-level commandlet with all those cahnges - I'll close this as it seems like you're already underway. Thanks!

PR #36, opened 2016-07-07
Invoke nunit3forassembly

nicolasvan:
Alternative PR to #35 

Since the command line parameters for NUnit v3 are quite different from v2.

### DIFF
--- a/Private/NUnit/3/Get-NUnit3ConsoleExePath.ps1
+++ b/Private/NUnit/3/Get-NUnit3ConsoleExePath.ps1
@@ -4,9 +4,13 @@ function Get-NUnit3ConsoleExePath {
         # The version of the nuget package containing the NUnit executables (NUnit.Console)
         [string] $NUnitVersion
     )
-    
+
+    if(!$NunitVersion.StartsWith('3.')) {
+        throw "Unexpected NUnit version '$NUnitVersion'. This function only supports Nunit v3"
+    }
+
     Write-Verbose "Using NUnit version $NUnitVersion"
     $NUnitFolder = Install-NUnitPackage $NUnitVersion
 
-    "$NUnitFolder\tools\nunit3.console.exe" | Resolve-Path
+    "$NUnitFolder\tools\nunit3-console.exe" | Resolve-Path
 }
--- a/Private/NUnit/3/Get-NUnit3ConsoleExePath.ps1
+++ b/Private/NUnit/3/Get-NUnit3ConsoleExePath.ps1
@@ -1,0 +1,12 @@
+function Get-NUnit3ConsoleExePath {
+    [CmdletBinding()]
+    param(
+        # The version of the nuget package containing the NUnit executables (NUnit.Console)
+        [string] $NUnitVersion
+    )
+    
+    Write-Verbose "Using NUnit version $NUnitVersion"
+    $NUnitFolder = Install-NUnitPackage $NUnitVersion
+
+    "$NUnitFolder\tools\nunit3.console.exe" | Resolve-Path
+}
--- a/Private/NUnit/3/Get-NUnit3ConsoleExePath.ps1
+++ b/Private/NUnit/3/Get-NUnit3ConsoleExePath.ps1
@@ -12,5 +12,11 @@ function Get-NUnit3ConsoleExePath {
     Write-Verbose "Using NUnit version $NUnitVersion"
     $NUnitFolder = Install-NUnitPackage $NUnitVersion
 
-    "$NUnitFolder\tools\nunit3-console.exe" | Resolve-Path
+    $nunit3ActualPackageName = "NUnit.ConsoleRunner"
+    if ($NUnitVersion.StartsWith("3.0") -or $NUnitVersion.StartsWith("3.1")) {
+      $nunit3ActualPackageName = "NUnit.Console"
+    }
+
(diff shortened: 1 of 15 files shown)


nicolasvan:
@GrahamTheCoder, @nyctef, are you guys keen to review this? I think it's getting there...

`Invoke-NUnit3ForAssembly` which uses `Build-NUnit3CommandLineArguments` and `Get-NUnit3ConsoleExePath`.

trading some code duplication vs better clarity :smile:

Private/NUnit/3/Get-NUnit3ConsoleExePath.ps1
```diff
@@ -0,0 +1,16 @@
...
+    )
+
+    if(!$NunitVersion.StartsWith('3.')) {
+        throw "Unexpected NUnit version '$NUnitVersion'. This function only supports Nunit v3"
+    }
+
+    Write-Verbose "Using NUnit version $NUnitVersion"
+    $NUnitFolder = Install-NUnitPackage $NUnitVersion
+
+    "$NUnitFolder\tools\nunit3-console.exe" | Resolve-Path
```
GrahamTheCoder: Is there an x86 version?

GrahamTheCoder: Ah, nvm, found it's now passed as a parameter :)

GrahamTheCoder:
I've done a quick skim of the file changes and it seems like what I was expecting - I'll give it a manual test on a branch of SqlClone and report back

GrahamTheCoder:
Yeah the test running seemed to work well on [this build](http://buildserver.red-gate.com/viewLog.html?buildId=7485902&buildTypeId=SQLClone_Build&tab=buildResultsDiv). It failed trying to merge the coverage reports but I think that's just because it accidentally recurses into node modules which is why we have this method in our build script:
https://github.com/red-gate/SqlClone/blob/master/.build/build.ps1#L19

nicolasvan:
Sweet :smile: 
yeah I've been very cheeky about that one.
Committed today : https://github.com/red-gate/RedGate.Build/commit/e878c084ce00341f0ae3d84fe48b0a4a6bcbf0ab

_basically ignore it_... :worried:

GrahamTheCoder:
I rebased onto master.

Now I've done that I've noticed that the test results aren't being imported - though I might have set up something wrong in our build script:
https://github.com/red-gate/SqlClone/pull/91
http://buildserver.red-gate.com/viewLog.html?buildId=7486913&buildTypeId=SQLClone_Build
The only tests shown there are the ones run by pester, not by NUnit

nicolasvan:
I did remove the manual import of the test results to Teamcity since NUnit 3 has a `--teamcity` parameter... Maybe this does not work as I expected? How about when trying with the latest version of NUnit? (3.4+ I believe...)

Get Outlook for Android<https://aka.ms/ghei36>



On Thu, Jul 7, 2016 at 5:51 PM +0100, "GrahamTheCoder" <notifications@github.com<mailto:notifications@github.com>> wrote:


I rebased onto master.

Now I've done that I've noticed that the test results aren't being imported - though I might have set up something wrong in our build script:
red-gate/SqlClone#91<https://github.com/red-gate/SqlClone/pull/91>
http://buildserver.red-gate.com/viewLog.html?buildId=7486913&buildTypeId=SQLClone_Build
The only tests shown there are the ones run by pester, not by NUnit

-

GrahamTheCoder:
Ok trying that here: http://buildserver.red-gate.com/viewLog.html?buildId=7487182&tab=buildResultsDiv&buildTypeId=SQLClone_Build

GrahamTheCoder:
Hmmm teamcity importing still doesn't seem to be working http://buildserver.red-gate.com/viewLog.html?buildId=7487251&buildTypeId=SQLClone_Build&tab=buildLog

nicolasvan:
:disappointed: 

Maybe let's just revert this commit https://github.com/red-gate/RedGate.Build/commit/698d336b7dd797663b8b638b4efc9043bf735f42 
(don't use `--teamcity` and import the test results from xml instead)

We can always revisit this later once we learn a little bit more how NUnit 3 actually work?

GrahamTheCoder:
Yeah, probably makes sense

GrahamTheCoder:
Hmm, no luck even with that reverted
http://buildserver.red-gate.com/viewLog.html?buildId=7487506&buildTypeId=SQLClone_Build&tab=buildLog

nicolasvan:
I pushed another change, completely untested !
I suspect that TC does not support parsing the XML result file outputted by nunit3.

so let's try to output them in the nunit 2 format?
based on that: https://groups.google.com/forum/#!topic/nunit-discuss/x7BhICiYQE0

GrahamTheCoder:
The version we have did have [some level](https://blog.jetbrains.com/teamcity/2016/01/teamcity-9-1-6-is-out-now/) of NUnit3 support, but yeah worth a try

nicolasvan:
So in the end, what it took to get this to work with Teamcity was to install the `NUnit.Runners` package (https://www.nuget.org/packages/NUnit.Runners/). It brings down `NUnit.ConsoleRunner` as well as `NUnit.Extension.TeamCityEventListener` which is the extensions that is needed for Teamcity integration :sweat_smile: 

I think this should be OK to merge and you guys can start using this :smile:

PR #39, opened 2016-08-11
Add Expand-ZipArchive cmdlet

alastairs:
Expose under the alias Unzip-Files to mirror the Zip-Files alias.


nicolasvan:
Looks fine to me :). I pushed a change so that we don't use try catch. 
if `$global:ErrorActionPreference -ne 'Stop'`, exceptions are not raised and can't really be caught or something....

alastairs:
Looks like either there are redundant parentheses, or the first closing parenthesis is in the wrong place...

PR #40, opened 2016-08-17
Enable code coverage

richardjm:
Doesn't seem to be a breaking change but conforms to docs now for `-EnableCodeCoverage` as switch not argument.

I tested `-EnableCodeCoverage $true` and it still worked.

### DIFF
--- a/Public/Invoke-NUnit3ForAssembly.ps1
+++ b/Public/Invoke-NUnit3ForAssembly.ps1
@@ -52,7 +52,7 @@ function Invoke-NUnit3ForAssembly {
 
   $AssemblyPath = Resolve-Path $AssemblyPath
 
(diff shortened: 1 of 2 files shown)


nicolasvan:
Using `-EnableCodeCoverage $true` doesn't actually pass `$true` to the $EnableCodeCoverage parameter.
`-EnableCodeCoverage` on its own is enough. The value of `$true` actually gets passed to the next parameter (if any) 

So using `-EnableCodeCoverage $false` will keep code coverage enabled. (quick local test confirms that)

The proper way to set it using a boolean would be to use: `-EnableCodeCoverage:$true` or `-EnableCodeCoverage:$false`

So that's a breaking change, let's not do it.

PR #43, opened 2016-09-08
Invoke nunit fixes

nicolasvan:
@angellaa This is to fix the breaking change that is breaking your build.

(basically a new private function made `-TestResultFilenamePattern` a required parameter which it wasn't before)

I added a clunky test so that we get to run "Invoke-NUnitForAssembly" when we build `RedGate.Build`. It's ugly but hey this might catch things before we break your world again :blush:


angellaa:
Cool thanks Nico.

PR #45, opened 2016-09-28
Don't touch line endings ever!

rmc47:
Git's CRLF is evil. Source control should never be touching file contents.


nicolasvan:
Sorry I don't think we should merge this.

* This might mess up with the files that are already committed?
* I would rather go with the default settings offered by github. 
 * I think this `.gitattributes` was created by github when I first created that repo. (might be the default `VisualStudio` one?) 
 * https://help.github.com/articles/dealing-with-line-endings/ seems to suggest `git config core.autocrlf` to be set to true.
 * I believe `git config core.autocrlf` is set to true by default on windows (when installing git). 
 * As for this repo, if this powershell cross platform thing takes off, we're already set to hack on linux 😄 



> Git's CRLF is evil. Source control should never be touching file contents.

That's an opinion, maybe a valid opinion. But git does it and it's fine for most use case unless opinionated developers start changing the default. (If the git installer didn't give us the option to choose, we might not even be aware and life would be good)

===
As a side note:

I've been there when I started using git and ended up the only developer changing the default. I felt so right for doing it at the time, but no-one else cared. (*How could they not care! We were a 100% Microsoft shop, it felt crazy to commit files with LF only. Did they not read the options when they installed git interactively for the first time?*). Well ultimately, after I messed up the company's repos, I had to backtrack, we all kept the default, wrote some fucking software, life was good again.

I suggest we do the same here 😅

rmc47:
I think the problem is we already *have* a mix of people's global options. As far as I'm aware, you can't use gitattributes to force autocrlf back to on if someone has it off on a PC, whereas you can force it off if someone has it on, if that makes sense.

So the setting that guarantees everything is consistent is to force it off using gitattributes...

nicolasvan:
https://help.github.com/articles/dealing-with-line-endings/ suggests otherwise 😄 

> Optionally, you can configure the way Git manages line endings on a per-repository basis by configuring a special *.gitattributes* file. This file is committed into the repository and overrides an individual's core.autocrlf setting, ensuring consistent behavior for all users, regardless of their Git settings.

rmc47:
I stand corrected :-). In which case, the important thing is all our of our repos *have* such a file, and the decision on its contents is less important...

So yup, this can stay closed.

PR #46, opened 2016-09-28
Advanced releasenotes

richardjm:
New Select-ReleaseNotes for much more advanced RELEASENOTES.md parsing, also has a strapline accumulator for check for updates releases.

Changed behaviour of Read-ReleaseNotes to fail if the n-part version isn't the first encountered (was seeing issues in some RELEASENOTES finding a much lower version that searched due to incorrect match far down the file)


richardjm:
Thanks, did wonder about the script: thing, I was trying to make those functions "private".

PR #47, opened 2016-10-05
Allows dots in strap line

richardjm:
Dots weren't allows to try to prevent "100. Rubbish" which github would render as "1. Rubbish" - now force a dash(-) between the priority and the strap line text instead.

### DIFF
--- a/Public/Select-ReleaseNotes.ps1
+++ b/Public/Select-ReleaseNotes.ps1
@@ -72,7 +72,7 @@ function GetSummary($productName, $release, $accumulator) {
   Date = '^#+\s*.*(?<date>\d\d\d\d.\d\d.\d\d)'
   Header = '^#+\s*(?<header>.+):?$'
   StraplineBlock = '^#+\s*Strapline\s*$'
-  Strapline = '^\s*(?<priority>[0-9]+)\s*-?\s*(?<feature>[^\.]*)\s*$'
+  Strapline = '^\s*(?<priority>[0-9]+)\s*-\s*(?<feature>.*)\s*$'
 
   Example input
   -------------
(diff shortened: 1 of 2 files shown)


Public/Select-ReleaseNotes.ps1
```diff
@@ -190,7 +190,7 @@ function Select-ReleaseNotes {
     $HeaderRegex = '^#+\s*(?<header>.+):?\s*$'
     $DateRegex = '^#+\s*.*(?<date>\d\d\d\d.\d\d.\d\d)'
     $StraplineStartRegex = '^#+\s*Strapline\s*$'
-    $StraplineRegex = '^\s*(?<priority>[0-9]+)\s*-?\s*(?<feature>[^\.]*)\s*$'
+    $StraplineRegex = '^\s*(?<priority>[0-9]+)\s*-\s*(?<feature>.*)\s*$'
```
ChrisLambrou: This is a breaking change. Are we sure that builds with release notes that don't currently contain the mandatory dash won't silently fail? i.e. do builds blow up if the new format isn't satisfied, or will the strapline simply be absent from those builds?

ChrisLambrou: Also, this isn't necessarily a bug introduced by this PR, but it seems as though the intention of the regex is to exclude trailing white-space before the end of the line, right? The regex ends with `\s*$`, which lies after the `<feature>` group. However, both the new `(?<feature>.*)` and the old `(?<feature>[^\.]*)` would match greedily, which would include that trailing white-space. Maybe your new regex could end with `(?<feature>.*?)\s*$` instead?

ChrisLambrou:
It looks reasonable to me, as long as you've considered how to handle projects that might get broken by this change.

richardjm:
So far this facility isn't used, I was just testing the first use of it. So a breaking change but no uses so far.

ChrisLambrou:
:shipit:

PR #49, opened 2016-11-11
Try passing through `where` argument

GrahamTheCoder:
Need to use this in SQL Clone


GrahamTheCoder:
Seems to exit with code 1.

* [ ] Write a test and make it pass

GrahamTheCoder:
I'm giving up on this for now - I don't have time to do it properly

PR #50, opened 2016-12-20
Add Dot Cover Process FIlter Argument

PeterGerrard:
Add argument so we can pass `/ProcessFilter` argument to dotcover (was added in 2016.2)

Command line reference: https://www.jetbrains.com/help/dotcover/2016.3/dotCover__Console_Runner_Commands.html#cover

We want this to filter out `sqlservr.exe` from our coverage


nicolasvan:
LGTM

PR #52, opened 2017-02-27
Return blocks in order from Select-ReleaseNotes

ChrisHurley:
`Select-ReleaseNotes` currently stores each block in the release notes (under a header, e.g. "Features", "Fixes" etc.) indexed as an object property. This is a dictionary, and when release scripts call `GetEnumerator()` on it, the order the blocks is returned is not specified. It appears to be deterministic for a given set of inputs, and returns "General", "Features", "Fixes" in the insertion order, but the addition of another header called "Introduction" messes up the ordering. This means that our release notes were being jumbled up.

This PR makes the `Blocks` property an array of name/value objects rather than an object with properties. As new members are added to the array as the sections are encountered in the release notes, this should mean that the sections are ordered as they first appeared in the input. As far as I can see, all the release scripts that use `Select-ReleaseNotes` call `GetEnumerator()` on `Blocks` and then access its members using `.Name` and `.Value`, so I believe that this won't break any existing consumers, unless they were relying on the previous unspecified ordering.

I have, however, had to update the tests to work with this. I've added an additional test with the `Introduction` case added, and also one that validates that the array can still be accessed with `GetEnumerator()`.

### DIFF
--- a/Public/Select-ReleaseNotes.ps1
+++ b/Public/Select-ReleaseNotes.ps1
@@ -212,7 +212,6 @@ function Select-ReleaseNotes {
     $Accumulator = @{}
     $StraplineAccumulator = $false
     $Release = $nul
-    $CurrentBlockIndex = 0;
     
     foreach($Line in $Lines) {
         $Line = $Line.Trim()
--- a/Public/Select-ReleaseNotes.ps1
+++ b/Public/Select-ReleaseNotes.ps1
@@ -28,19 +28,8 @@ function TrimBlocks($release) {
     } 
 }
 
-function FinalizeRelease($release, [string]$productName, $accumulator, $convertBlocksToObject) {
+function FinalizeRelease($release, [string]$productName, $accumulator) {
     TrimBlocks($release)
-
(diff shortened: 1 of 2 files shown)


samblackburn:
Random aside: The [Compare release notes](https://documentation.red-gate.com/display/SC12/SQL+Compare+12.1+release+notes) have a lot of versions listing `Fixes` before `Features` which is kinda sad.  I'm not sure whether the best solution is to reorder them during the release or just be more disciplined about how we enter release notes.

ChrisHurley:
@samblackburn Is that because you wanted them to be in that order, or by mistake?

samblackburn:
I think people were just appending stuff to the `NEXT_COMPARE_RELEASENOTES.md` so if the first piece of work done was a bugfix, the heading ended up at the top of the file :(

samblackburn:
Cool, I've verified that: 
 - `Select-ReleaseNotes` didn't get a new switch after all
 - All consumers call `$notes.Blocks.getEnumerator()` and then call `$note.Name` and `$note.Value`

PR #60, opened 2017-03-10
Allow Update-RedgateNugetPackages to update all packages

nicolasvan:
By passing `-IncludedPackages '*'` and not passing in `-ExcludedPackages`

This seems to be actually needed by some of our repos 😞


nicolasvan:
or maybe not needed, hmmm not so sure anymore...

PR #62, opened 2017-03-15
[breaking] Remove -DoNotCommitChanges parameter from Update-RedgateNugetPackages 

nicolasvan:
Instead make `-GithubApiToken` optional.

* if `-GithubApiToken` is set, we commit/push changes and create/update a PR
* if `-GithubApiToken` is not set, we do not commit changes

This is going to break any of our build that use `-DoNotCommit` but hey, this should make things simpler?


allymparker:
Yeah this occurred to me last night!  You'll need to change the SharedSSMS build scripts though too :/

nicolasvan:
yeah hopefully, this should get more stable after that :P

nicolasvan:
shipping as -DotNotCommitChanges is currently broken anyway.

PR #63, opened 2017-03-20
Commit a custom-built nuget.exe...

nicolasvan:
This was built from https://github.com/NuGet/NuGet.Client/releases/tag/3.5.0-rtm + https://github.com/NuGet/NuGet.Client/pull/1248

Clearly not many people are using `nuget update` with nuget packages that include `.targets` files 😥

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -98,15 +98,13 @@ try {
   }
 
   # Download NuGet if necessary.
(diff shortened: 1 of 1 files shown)


timdhughes:
We would be using nuget update with targets if it weren't broken!

PR #64, opened 2017-03-28
Allow user to specify a package version override

samblackburn:
This is needed when a package has a x.y.z version but isn't actually semvered, e.g. NUnit

 - [ ] Try this In SQL Compare Engine and see if it fixes the nuget update process


nyctef:
Isn't this already covered by the `ExcludedPackages` option?

samblackburn:
That allows you to specify the correct nuget dependencies, but the nuspec updating ignores it

nicolasvan:
This is going to be useful, thanks! 👍

samblackburn:
Wait, I changed my mind!

PR #70, opened 2017-09-16
Added Test-NugetPackage cmdlet

ChrisLambrou:
Cmdlet used to determine whether or not a specific version of a package exists.

Normally one would use the `nuget.exe list` command, but there's [a long-standing unfixed bug](https://github.com/NuGet/NuGetGallery/issues/3274) that prevents this. This cmdlet provides an alternative method for detecting the presence or absence of a specific package, without having to resort to manually invoking `nuget.exe list`, scraping the output, then realising it's buggy.

The current implementation simply invokes `Install-Package` and then checks to see whether the operation succeeds or fail. Whilst the Pester tests passed, unfortunately this triggered TeamCity build failures because the console error output produced by `nuget.exe` when checking for a non-existent package was interpreted as a sign of a build failure. To work around this, I've modified `Install-Package` to provide a `-Silent` parameter that suppresses error output.

### DIFF
--- a/Public/Nuget/Install-Package.ps1
+++ b/Public/Nuget/Install-Package.ps1
@@ -9,6 +9,8 @@
         The name/id of the nuget package to install.
         .PARAMETER Version
         The version of the nuget package to install.
+        .PARAMETER Silent
+        Suppresses stdout and stderr output when invoking NuGet.
         .OUTPUTS
         A string that is the full path of the folder that the package
         was installed to.
@@ -22,7 +24,9 @@ function Install-Package
         [string] $Name,
 
         [Parameter(Mandatory = $true)]
-        [string] $Version
+        [string] $Version,
+
+        [switch] $Silent
     )
 
     # Looks for an existing package that matches the requested id and version.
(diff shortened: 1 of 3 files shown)


nyctef:
Ah, I see. I guess the way we usually get around that is to increment the third number in the version for every build and just publish every package that hits `master` (assuming there's probably something worth publishing).

I'm happy for this to be merged but I don't know if anyone else is going to use it

ChrisLambrou:
Regarding your simpler version, I did think about using the same mechanism. It falls down when there are multiple package sources, though.

Next thoughts was to enumerate through all the package sources, and check them one by one. For http(s) packages sources, the web-request technique is fine. For file-base package sources, it gets more complicated, as the package can appear in two different places. It's doable, but there's more code to go wrong here.

Next thought was to just attempt to install the package, and rely on NuGet's built-in behaviour of searching all of the package sources for the package. I had thought of installing the package to a temporary folder, so it could be deleted afterwards. However, I opted to just use our current `Install-Package` cmdlet. It's a trade-off between having a simpler implementation at the cost of polluting the local `\.build\packages\` folder. Yeah, it's a little odd that it pollutes the local workspace, but I don't think there's any harm in doing this.

ChrisLambrou:
I want to avoid publishing every package because (a) I don't want to spam the `nuget.org` package source (I'm hoping to make the packages public), and (b) for any given build on a branch, the majority of packages will be unchanged (only the library being worked on would be changed). Yes, in theory each package should live in its own repo and/or have its own build process. In practice, that would make each package too heavyweight. This is for the source-only micro-services repo I posted in `eng-general` a few months ago, and I'm really striving for them to be as lightweight as possible.

ChrisLambrou:
> but I don't know if anyone else is going to use it

Well I only have [one use for it](https://github.com/ChrisLambrou/MicroLibraries/blob/master/.build/build.ps1#L193) at the moment, but I'm hoping to use it for a bunch of sanity checks in some of the COREDev builds, where I'd like the builds to fail early.

nyctef:
> It's a trade-off between having a simpler implementation at the cost of polluting the local \.build\packages\ folder. Yeah, it's a little odd that it pollutes the local workspace, but I don't think there's any harm in doing this.

Oh, I didn't realise it would put stuff into `\.build\packages` (instead of actually installing packages into projects in the solution and causing uncommitted changes) - that does seem alright

Thanks for explaining everything - I think I understand how this is useful now :)

PR #81, opened 2017-12-05
Upgrade Invoke-Build to 4

mark-raymond:
Looking at the [Invoke-Build release notes](https://github.com/nightroman/Invoke-Build/blob/v4.0.0/Release-Notes.md#v400), nothing that significant happened at 4.0.0, so we should just upgrade.

(Also, now [my PR to Invoke-Build](https://github.com/nightroman/Invoke-Build/pull/107) is merged, we need to upgrade Invoke-Build for me to benefit from it :smile:)

### DIFF
--- a/paket.dependencies
+++ b/paket.dependencies
@@ -1,5 +1,5 @@
 source https://nuget.org/api/v2
 
 nuget Nuget.CommandLine ~> 4
-nuget Invoke-Build ~> 3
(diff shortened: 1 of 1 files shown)


nightroman:
Yes, I recommend upgrade, it is safe for most users and brings only improvements :)
As for the PR, I will make some changes and ask to review and try, if you do not mind.

PR #83, opened 2018-01-19
Always generate dotCover HTML report

Greg-Smulko:
Always generate dotCover HTML report.
The reason for this change is that SonarQube requires it for proper integration.

### DIFF
--- a/Public/Merge-CoverageReports.ps1
+++ b/Public/Merge-CoverageReports.ps1
@@ -44,11 +44,7 @@ function Merge-CoverageReports {
   Write-Verbose "Merging snapshots: `r`n$snapshots`r`n`r`nto $MergedSnapshotPath"
 
   & $DotCoverPath merge /Source="$($snapshots -join ';')" /Output="$MergedSnapshotPath"
-
-  if( $env:TEAMCITY_VERSION -eq $null ) {
-    # Create an HTML report if running outside of Teamcity to help with debugging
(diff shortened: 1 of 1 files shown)


Public/Merge-CoverageReports.ps1
```diff
@@ -45,7 +49,7 @@ function Merge-CoverageReports {
 
   & $DotCoverPath merge /Source="$($snapshots -join ';')" /Output="$MergedSnapshotPath"
 
-  if( $env:TEAMCITY_VERSION -eq $null ) {
+  if( $env:TEAMCITY_VERSION -eq $null -or $GenerateHtmlReport -eq $True) {
```
nicolasvan: Since we're likely to use sonarqube more and more, how about always generating the HTML report?
In that case, we could remove that condition and the `GenerateHtmlReport` parameter and keep things simpler?

Greg-Smulko: I'm happy to do it.
I just chose a safer path, as I wasn't sure if that may increase the build time for other teams perhaps?

nicolasvan: Let's give it a go and see what happens? Hopefully it's not slow enough to be noticeable... If things really slow down, I can bring it back 😄

Greg-Smulko: Done. :)

PR #90, opened 2018-03-22
Add support for new csproj in update package

PeterGerrard:
Adds support for new csproj format when updating packages.

Requires being able to run dotnet tools against any csproj using the new `PackageReference` syntax.

This will only run if there is a `PackageReference` so shouldn't affect most tools. I tested it against DLM Automation. Also prevents failure if there are no package configs in the repo

### DIFF
--- a/Public/Update-RedgateNugetPackages.ps1
+++ b/Public/Update-RedgateNugetPackages.ps1
@@ -183,11 +183,11 @@ function UpdatePackageReferences([string]$RootDir, [string[]]$IncludedPackages, 
     Get-ChildItem $RootDir -Filter *.csproj -Recurse | % { 
         $file = $_.FullName
         $packages = @()
-        (Get-Content $_.FullName | % {
-            if ($_  -match "<PackageReference Include=`"(?<PackageName>[\.\w]+)`" Version=`"(?<PackageVersion>[\d\.]+)`"") {
-                $packages = $packages + @(@{ Name = $Matches.PackageName; Version = $Matches.PackageVersion })
-            }
-        })
+        $xml = [xml](Get-Content $_.FullName)
+
+        (Select-Xml "//PackageReference" $xml).Node | % {
+            $packages = $packages + @(@{ Name = $_.Include; Version = $_.Version })
+        }
         
         $filteredPackages = @()
         foreach($pattern in $IncludedPackages) {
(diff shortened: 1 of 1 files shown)


PeterGerrard:
We can put a new switch in to allow new csproj updates so we can test it in live for a few weeks without turning it on by default. Opinions?

PeterGerrard:
@samblackburn could you take a look?

Public/Update-RedgateNugetPackages.ps1
```diff
@@ -169,3 +150,76 @@ function GetNugetPackageIds(
 
     return $FilteredPackageIDs | Select-Object -Unique | Sort-Object
```
samblackburn: Might be worth moving the `Unique` operation earlier, so the package count etc. are accurate

Public/Update-RedgateNugetPackages.ps1
```diff
@@ -68,28 +68,9 @@ Function Update-RedgateNugetPackages
...
-                Select-Object -ExpandProperty Path |
-                Update-NuspecDependenciesVersions `
-                    -PackagesConfigPaths $packageConfigFiles.FullName `
-                    -DoNotUpdate $ExcludedPackages `
-                    -Verbose
-        }
-        
-        $UpdatedPackages = $UpdatedPackages | Select -Unique
-        Write-Output $UpdatedPackages
+        $UpdatedPackages += UpdatePackageReferences -RootDir $RootDir -IncludedPackages $IncludedPackages -ExcludedPackages $ExcludedPackages
```
samblackburn: Do we need to dedupe `$UpdatedPackages` after appending these values, or is it a set?

PeterGerrard: Not a set, we should dedupe (even though their can't currently be duplicates between the two)

I also noticed I don't dedupe leaving UpdatePackageReferences

Public/Update-RedgateNugetPackages.ps1
```diff
@@ -169,3 +150,76 @@ function GetNugetPackageIds(
...
+    
+    $UpdatedPackages | Select -Unique | Write-Output
+}
+
+function UpdatePackageReferences([string]$RootDir, [string[]]$IncludedPackages, [string[]]$ExcludedPackages) {
+    Get-ChildItem $RootDir -Filter *.csproj -Recurse | % { 
+        $file = $_.FullName
+        $packages = @()
+        (Get-Content $_.FullName | % {
+            if ($_  -match "<PackageReference Include=`"(?<PackageName>[\.\w]+)`" Version=`"(?<PackageVersion>[\d\.]+)`"") {
```
samblackburn: Surprised a regex is easier to use than XML querying here, but ok

PeterGerrard:
@samblackburn made those changes, the xml is both simpler than the regex and works in a more circumstances. So I moved to it.

samblackburn:
:+1: Looks good - I suppose in a perfect world it would be nice to have a test around this, but it would either be very artificial or very fragile

PR #91, opened 2018-03-23
Fix various package auto update builds

samblackburn:
## What does this PR do?
Uses TLS 1.2 when auto-creating pull requests, putting it back again after each request

## Why?
Web requests to Github fail unless TLS 1.2 is used

## Why put it back? Why not just set it globally?
We're worried that product integration issues might be hidden if we set the test environment up so that such requests will just work.

## How has it been checked?
Mainly TDD

## How has it been tested?
 - [ ] Kick off an auto package update build pinned to this prerelease version, I guess?


Public/GitHub/Use-Tls.ps1
```diff
@@ -0,0 +1,27 @@
...
+        Sets the encryption provider to use TLS 1.2, putting it back again once the given code block has exited or thrown
+        .PARAMETER delegate
+        A code block (typically involving one or more web requests) to be executed
+#>
+#requires -Version 2
+function Use-Tls
+{
+    [CmdletBinding()]
+    param(
+        [Parameter(Mandatory = $true)]
```
PeterGerrard: It would be nice to type this: `[ScriptBlock]`

Public/GitHub/Use-Tls.ps1
```diff
@@ -0,0 +1,27 @@
...
+        $delegate
+    )
+    Try {
```
PeterGerrard: PowerShell keywords should be lower case: https://github.com/PoshCode/PowerShellPracticeAndStyle/blob/master/Style-Guide/Code-Layout-and-Formatting.md

At the very least we should be consistent

Tests/Use-Tls.Tests.ps1
```diff
@@ -0,0 +1,9 @@
+#requires -Version 2 -Modules Pester
+
+Describe 'Use-Tls' {
```
PeterGerrard: I think a test that checks the tls settings are correct whilst running might be useful. Check 
```PowerShell
[Net.ServicePointManager]::SecurityProtocol -band [Net.SecurityProtocolType]::Tls12 } } | Should -Be [Net.SecurityProtocolType]::Tls12
```

_band_ is the bitwise and operator

PR #94, opened 2018-04-16
Add support to Invoke-SigningService for signing libs in NuGet packages

ChrisLambrou:
**What does this PR do?**

Adds support for calling `Invoke-SigningService` with a NuGet package. The cmdlet then signs any assembly dlls within the `libs` sub-folder of the package.

**Why?**

To date, many build scripts contain the following three separate steps.

1. Compile a solution.
2. Invoke the signing service on specific assembly dlls that were built in step 1.
3. Generate the NuGet packages based on manually maintained `.nuspec` files.

With a gradual move to .NET Standard, .NET Core, the new VS 2017 project format, and multi-platform targeting, steps 1 and 3 are now carried out at the same time by msbuild and the dotnet cli tools. This no longer leaves us with a clean opportunity to sign the assembly dlls after they've been created, but before they've been packaged into a NuGet package.

This new functionality allows us to perform the assembly signing after msbuild has done its job of building and packaging them.

**What alternatives were considered?**

- Hook into the msbuild process and invoke the signing after the assemblies have been built, but before they're packaged into a NuGet package. Rejected because this feels a bit too much like the old build magic. It's overly intrusive into the build process, and potentially slows the builds during development.

- Update the signing service so that it can sign the NuGet packages, whereby it would unpack the NuGet package, sign the assemblies, then rebuild the package. Rejected mainly because performing this work on the client means it's more transparent to our users (i.e. devs in the product division). For example, if someone decides that they don't want all assemblies in the `lib` folder to be signed, changing the behaviour in the client is more straightforward than chaning the server.

**How's it tested?**

Manually. The existing tests for `Invoke-SigningService` aren't particularly comprehensive. Maybe we should explore ways to make this more testable?

Update: I've retested this manually after the code review changes. All seems good. Would still be nice to craft some automated checks. Will save than for another PR.

### DIFF
--- a/Public/Invoke-SigningService.ps1
+++ b/Public/Invoke-SigningService.ps1
@@ -188,7 +188,7 @@ function Invoke-SigningServiceForNuGetPackage {
             ForEach-Object {
                 Write-Verbose "Signing $($_.FullName.Substring($ContentsDir.Length + 1))"
                 if ($HashAlgorithm) {
-                    $Null = Invoke-SigningService `
+                    Invoke-SigningService `
                         -FilePath $_.FullName `
                         -SigningServiceUrl $SigningServiceUrl `
                         -Certificate $Certificate `
@@ -197,7 +197,7 @@ function Invoke-SigningServiceForNuGetPackage {
                         -MoreInfoUrl $MoreInfoUrl `
                         -Force:$Force.IsPresent
                 } else {
-                    $Null = Invoke-SigningService `
+                    Invoke-SigningService `
                         -FilePath $_.FullName `
                         -SigningServiceUrl $SigningServiceUrl `
                         -Certificate $Certificate `
--- a/Public/Invoke-SigningService.ps1
+++ b/Public/Invoke-SigningService.ps1
@@ -74,84 +74,132 @@ function Invoke-SigningService {
 
     process {
         # Simple error checking.
-        if ([String]::IsNullOrEmpty($SigningServiceUrl)) {
+        if ([string]::IsNullOrEmpty($SigningServiceUrl)) {
             throw 'Cannot sign file. -SigningServiceUrl was not specified and the SigningServiceUrl environment variable is not set.'
         }
         if (!(Test-Path $FilePath)) {
             throw "File not found: $FilePath"
         }
 
-        # Only sign the file if it does not already have a valid Authenticode signature
-        if(!$Force.IsPresent -and (Get-AuthenticodeSignature $FilePath).Status -eq 'Valid') {
-            Write-Verbose "Skipping signing $FilePath. It is already signed"
-            return $FilePath
-        }
-
-        # Determine the file type.
-        $FileType = $Null
-        switch ([IO.Path]::GetExtension($FilePath)) {
-            '.exe' { $FileType = 'Exe' }
-            '.msi' {
-                $FileType = 'Exe'
-                #  msi files cannot be double signed at the moment.
-                #  so tell the signing service to use sha256
-                if(!$HashAlgorithm) { $HashAlgorithm = 'sha256' }
-            }
-            '.dll' { $FileType = 'Exe' }
-            '.vsix' {
-                $FileType = 'Vsix'
-                if(!$HashAlgorithm) {
-                    throw @'
-Cannot sign vsix package. -HashAlgorithm was not specified.
-Use sha1 if targeting VS 2013 and older. Use sha256 if targeting VS 2015+
-'@
-                }
-            }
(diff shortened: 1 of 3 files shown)


Public/Invoke-SigningService.ps1
```diff
@@ -143,6 +154,73 @@ Use sha1 if targeting VS 2013 and older. Use sha256 if targeting VS 2015+
...
+    {
+        # First copy the NuGet package to the temp dir, giving it a .zip extension to please the Expand-Archive cmdlet.
+        Write-Verbose 'Copying NuGet package to the working dir'
+        $ZipFilePath = "$TempDir\package.zip"
+        Copy-Item -Path $FilePath -Destination $ZipFilePath
+    
+        # Next extract the zip file to a folder.
+        Write-Verbose 'Expanding NuGet package contents to the working dir'
+        $ContentsDir = [string] (mkdir "$TempDir\Contents")
+        Expand-Archive -Path $ZipFilePath -DestinationPath $ContentsDir
```
PeterGerrard: This won't handle paths with spaces in them. When repackaged they will have `%20` instead of spaces.

Public/Invoke-SigningService.ps1
```diff
@@ -115,6 +115,17 @@ Use sha1 if targeting VS 2013 and older. Use sha256 if targeting VS 2015+
...
+            '.nupkg' {
+                Invoke-SigningServiceForNuGetPackage `
+                    -FilePath $FilePath `
+                    -SigningServiceUrl $SigningServiceUrl `
+                    -Certificate $Certificate `
+                    -HashAlgorithm $HashAlgorithm `
+                    -Description $Description `
+                    -MoreInfoUrl $MoreInfoUrl `
+                    -Force:$Force.IsPresent
+                return $FilePath
```
PeterGerrard: I would rather see this function call another for both files and packages. Not a fan of this mutual calling with early out.

Public/Invoke-SigningService.ps1
ChrisLambrou: Why? You obviously know something specific about this, but I don't. Can you explain what the problem is in enough detail for me to be able to address it?

Public/Invoke-SigningService.ps1
ChrisLambrou: I'm not sure I follow. Do you mean it would be good for this function to delegate to one of two other functions - one for the special handling of NuGet packages, and the other for all the previously cases?

Public/Invoke-SigningService.ps1
PeterGerrard: When you unzip a NuGet package rather than nuget install or similar then it doesn't handle space in paths correctly.

`content/file with spaces.txt` gets unpacked as `content/file%20with%20spaces.txt`

Public/Invoke-SigningService.ps1
PeterGerrard: That's what I mean.

```powershell
if ($IsPackage) {
 Invoke-PackageSign
}
else {
 Invoke-FileSign
}
```

Public/Invoke-SigningService.ps1
ChrisLambrou: **<TL;DR> Although the names of the extracted package files in the temp working may not always precisely match those of the original files, it's enough that the unpack/repack process is transparent w.r.t. the zip entry names, and that any file name changes don't affect the signing process.**

Okay, I've looked into this a bit more, and I'm not sure anything actually needs to change. If you have any original file that contains spaces (or indeed any other characters that ought to be url encoded), the NuGet pack command applies url encoding to the file name when it creates the NuGet package. So indeed, `content.file with spaces.txt` results in a zip entry named `content/file%20with%20spaces.txt`. That's "normal behaviour for NuGet packages.

Now, consider what this PR does to the NuGet package. It unpacks the package like a standard zip file, signs the assembly dlls in the `lib` sub-folder, then repacks them like a standard zip file. The upshot is that the unpack/repack process is transparent w.r.t the names of the zip entries, even though the names of some unpacked files in the temp working folder might differ in name from the original raw files (those on which the original NuGet package was based). But so what? 

The only potential problem I can see is that one of those assemblies might get renamed, and that's somehow important w.r.t the signing process. In mitigation:

1. Even if there are unicode characters in any assembly name that mean it gets escaped, we'd still be able to reliably detect the `.dll` extension (that bit wouldn't be escaped), and we'd still sign the assembly, even though its unpacked name doesn't match that of the original file.
2. The file name is never actually sent to the signing service, only its contents. So the different name should not affect the actually signing process in any way that matters here.

PR #98, opened 2018-04-19
Add switch to Update-NuspecDependenciesVersions to not use version ranges

RobClenshaw:
With semantic versioning, we generate a range of values when we update a nuspec file. For example, `[1.2.3, 2.0.0)`. This is how strict semantic versioning works.

[Our policy](https://info.red-gate.com/eng/development-policies/libraries/versioning-assemblies) is not a strict interpretation of semantic versioning and as such we sometimes want to enforce a specific version rather than a range (i.e. SSMS plugins and apps with shared components running within a single appdomain). This switch will not change the current behaviour but will give us an opportunity to force specific versions if we wish.

### DIFF
--- a/Private/Nuget/Get-DependencyVersionRange.ps1
+++ b/Private/Nuget/Get-DependencyVersionRange.ps1
@@ -14,7 +14,8 @@ function Get-DependencyVersionRange
     param (
         [Parameter(Mandatory = $True, Position = 0, ValueFromPipeLine = $True)]
         [ValidateNotNullOrEmpty()]
-        [string] $Version
+        [string] $Version,
+        [switch] $SpecificVersion
     )
 
     if($Version.Contains('-'))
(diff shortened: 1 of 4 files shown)


juliahayward:
Looks good to me - perhaps @samblackburn should have a look as well?

Private/Nuget/Get-DependencyVersionRange.ps1
```diff
@@ -28,7 +29,7 @@ function Get-DependencyVersionRange
     }
 
     $versionParts = $currentVersion.Split(".")
-    if ($versionParts.Length -eq 3) #http://wiki/Semantic_Versioning
+    if ($versionParts.Length -eq 3 -And -Not $SpecificVersion) #http://wiki/Semantic_Versioning
```
samblackburn: The comment is a dead link - do we up to date documentation somewhere?

RobClenshaw:
OK, I've added a parameter description to the script header.

PR #100, opened 2018-04-24
Don't trim release notes lines

Michael-Upton:
Leading whitespace can be significant in markdown (it's an indent). Instead of trimming every line in Select-ReleaseNote, modify the relevant Regexs to allow leading whitespace.

I need this so that our auto-uploading release notes in DLMAutomation can have indented sections.


PeterGerrard:
Could you add some tests for this: https://github.com/red-gate/RedGate.Build/blob/master/Tests/Select-ReleaseNotes.Tests.ps1

PR #101, opened 2018-06-08
Ensure we use process when using the pipeline

PeterGerrard:
When adding The SupportsPipeline you need to add process into the method body, otherwise it will run end which will have the last value rather than running foreach value.


Greg-Smulko:
@PeterGerrard , is it still relevant? If so, should release notes be added?

PeterGerrard:
@Greg-Smulko Yes its still relevant. Though I will rebase and add release notes. Would you be willing to review?

PR #104, opened 2018-09-07
Set the working directory for a test run

RobClenshaw:
This adds the optional `-TargetWorkingDirectory` parameter to `Invoke-NUnit3ForAssembly` and `Invoke-DotCoverForExecutable` so that you can specify the working directory for the test run.

I've added three new unit tests to check that `Invoke-NUnit3ForAssembly` calls things in the expected way.

### DIFF
--- a/Public/Invoke-DotCoverForExecutable.ps1
+++ b/Public/Invoke-DotCoverForExecutable.ps1
@@ -73,7 +73,7 @@ function Invoke-DotCoverForExecutable {
     $DotCoverArguments += "/TargetArguments=$EscapedTargetArguments"
   }
 
-  if (![string]::IsNullOrWhiteSpace($TargetWorkingDirectory)) {
+  if ($TargetWorkingDirectory) {
     $DotCoverArguments += "/TargetWorkingDir=`"$TargetWorkingDirectory`""
   }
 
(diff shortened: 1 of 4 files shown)


Public/Invoke-DotCoverForExecutable.ps1
```diff
@@ -68,6 +73,10 @@ function Invoke-DotCoverForExecutable {
     $DotCoverArguments += "/TargetArguments=$EscapedTargetArguments"
   }
 
+  if (![string]::IsNullOrWhiteSpace($TargetWorkingDirectory)) {
```
samblackburn: Minor: I think `if ($TargetWorkingDirectory)` would be more concise

Public/Invoke-DotCoverForExecutable.ps1
```diff
@@ -68,6 +73,10 @@ function Invoke-DotCoverForExecutable {
...
+    $DotCoverArguments += "/TargetWorkingDir=$TargetWorkingDirectory"
```
samblackburn: Does this need escaping at all?

samblackburn:
:+1: nice to see test coverage!

Some very minor comments, I don't really mind if you merge without addressing them

Public/Invoke-DotCoverForExecutable.ps1
RobClenshaw: I was just doing what the other string arguments were doing. I can change it if you like.

Public/Invoke-DotCoverForExecutable.ps1
RobClenshaw: Almost certainly. Oops.

PR #107, opened 2018-09-13
Use an xslt transform to remove ignored tests

RobClenshaw:
This PR uses an xslt transform to remove ignored tests. The advantage of this is that it uses streams and avoids loading the whole contents of the file into memory. This can prevent out of memory exceptions with large test runs.

Additionally, some result files were observed during TeamCity runs to still be open after the test run had allegedly finished. The method now waits for the lock on the file to be released before

### DIFF
--- a/Public/Remove-IgnoredTests.ps1
+++ b/Public/Remove-IgnoredTests.ps1
@@ -33,7 +33,8 @@ function Remove-IgnoredTests {
     $DestinationFilePath = $ExecutionContext.SessionState.Path.GetUnresolvedProviderPathFromPSPath($DestinationFilePath)
   }
 
-  $whereClause = "//test-suite[(@result='Ignored' or @label='Ignored') and (" + (($ReasonsIgnored | ForEach-Object { "reason/message='$_'" }) -join " or ") + ")]"
+  $reasonsIgnoredString = ($ReasonsIgnored | ForEach-Object { "reason/message='$_'" }) -join " or "
+  $whereClause = "//test-suite[(@result='Ignored' or @label='Ignored') and (" + $reasonsIgnoredString + ")]"
 
   $stringBuilder = [System.Text.StringBuilder]::new()
   $stringBuilder.AppendLine("<xsl:stylesheet version=`"1.0`" xmlns:xsl=`"http://www.w3.org/1999/XSL/Transform`">") | Out-Null
--- a/Public/Remove-IgnoredTests.ps1
+++ b/Public/Remove-IgnoredTests.ps1
@@ -34,16 +34,16 @@ function Remove-IgnoredTests {
   }
 
   $reasonsIgnoredString = ($ReasonsIgnored | ForEach-Object { "reason/message='$_'" }) -join " or "
-  $whereClause = "//test-suite[(@result='Ignored' or @label='Ignored') and (" + $reasonsIgnoredString + ")]"
 
-  $stringBuilder = [System.Text.StringBuilder]::new()
-  $stringBuilder.AppendLine("<xsl:stylesheet version=`"1.0`" xmlns:xsl=`"http://www.w3.org/1999/XSL/Transform`">") | Out-Null
-  $stringBuilder.AppendLine("<xsl:template match=`"@* | node()`"><xsl:copy><xsl:apply-templates select=`"@* | node()`"/></xsl:copy></xsl:template>") | Out-Null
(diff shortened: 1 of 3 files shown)


Public/Remove-IgnoredTests.ps1
```diff
@@ -17,8 +17,7 @@ function Remove-IgnoredTests {
 
     # A list of ignored reason messages.
     # Only tests with a ignored reason matching a string in this list will be removed
-    # wildcards '*' are supported
-    [string[]] $ReasonsIgnored = @('*'),
+    [string[]] $ReasonsIgnored = @(),
```
samblackburn: You've changed this line but not the comment above.  Are they still consistent?

Public/Remove-IgnoredTests.ps1
```diff
@@ -34,22 +33,45 @@ function Remove-IgnoredTests {
     $DestinationFilePath = $ExecutionContext.SessionState.Path.GetUnresolvedProviderPathFromPSPath($DestinationFilePath)
   }
 
-  Write-Verbose "Loading test results from $TestResultsPath"
-  $xml = [xml](Get-Content $TestResultsPath)
+  $whereClause = "//test-suite[(@result='Ignored' or @label='Ignored') and (" + (($ReasonsIgnored | ForEach-Object { "reason/message='$_'" }) -join " or ") + ")]"
```
samblackburn: Minor: Might be worth extracting a variable here, I find these nested/quoted brackets a bit confusing

Public/Remove-IgnoredTests.ps1
RobClenshaw: The comment is still consistent, we just don't allow wildcards any more.

Public/Remove-IgnoredTests.ps1
```diff
@@ -34,22 +33,46 @@ function Remove-IgnoredTests {
...
+    $totalSleep += 10
+    Start-Sleep -s 10
   }
 
-  Write-Verbose "Saving updated results to $DestinationFilePath"
-  $xml.Save( $DestinationFilePath)
-}
+  Write-Host "Moving temporary file to the specified location"
+  Move-Item -Path $tempFileName -Destination $DestinationFilePath -Force -Verbose
+  Write-Host "Removing the temporary directory"
```
samblackburn: This debug info can probably be `Write-Verbose` now that everything is working

Public/Remove-IgnoredTests.ps1
```diff
@@ -34,22 +33,46 @@ function Remove-IgnoredTests {
...
+  $totalSleep = 0
+  while (-not (Test-FileUnlocked -Path $TestResultsPath))
+  {
+    if ($totalSleep -ge 300)
+    {
+      throw [System.TimeoutException] "File $testResultsPath has been locked for more than 5 minutes."
     }
+    Write-Host "File $testResultsPath locked. Sleeping for 10 seconds."
+    $totalSleep += 10
+    Start-Sleep -s 10
```
samblackburn: Perhaps it's worth extracting a `Wait-FileUnlocked` method?

Public/Remove-IgnoredTests.ps1
```diff
@@ -34,22 +33,46 @@ function Remove-IgnoredTests {
...
-  $xml = [xml](Get-Content $TestResultsPath)
+  $reasonsIgnoredString = ($ReasonsIgnored | ForEach-Object { "reason/message='$_'" }) -join " or "
+  $whereClause = "//test-suite[(@result='Ignored' or @label='Ignored') and (" + $reasonsIgnoredString + ")]"
 
-  $ignoredTests = $xml | Select-Xml -XPath '//test-suite[@result="Ignored" or @label="Ignored"]'
+  $stringBuilder = [System.Text.StringBuilder]::new()
+  $stringBuilder.AppendLine("<xsl:stylesheet version=`"1.0`" xmlns:xsl=`"http://www.w3.org/1999/XSL/Transform`">") | Out-Null
+  $stringBuilder.AppendLine("<xsl:template match=`"@* | node()`"><xsl:copy><xsl:apply-templates select=`"@* | node()`"/></xsl:copy></xsl:template>") | Out-Null
+  $stringBuilder.AppendLine("<xsl:template match=`"$whereClause`" />") | Out-Null
+  $stringBuilder.AppendLine("</xsl:stylesheet>") | Out-Null
```
samblackburn: Would a here-string be more readable here? https://stackoverflow.com/a/26045894

PR #109, opened 2018-10-19
Workaround powershell 5.1.17763.1 throwing errors on 'git checkout -B' 

nicolasvan:
Looks like powershell 5.1.17763.1 (Windows Server 2019) is throwing errors even though `git checkout -B` is successful. Suspect that is because `git checkout -B` prints its output on stderr... 😓😓

### DIFF
--- a/Public/GitHub/Push-GitChangesToBranch.ps1
+++ b/Public/GitHub/Push-GitChangesToBranch.ps1
@@ -49,7 +49,9 @@ Function Push-GitChangesToBranch
     git config --local user.name $CommitUsername
     git config --local user.email $CommitEmail
 
-    execute-command { & git checkout -B $BranchName }
+    # Redirect git's stderr to stdout then to Write-Host
+    # for powershell v5.1.17763.1 (Windows Server 2019) to stop throwing errors
+    execute-command { & git checkout -B $BranchName 2>&1 | Write-Host }
     execute-command { & git add --all }
     execute-command { & git commit -am $CommitMessage }
     execute-command { & git push -f $RemoteName $BranchName`:$BranchName }


Public/GitHub/Push-GitChangesToBranch.ps1
```diff
@@ -49,7 +49,9 @@ Function Push-GitChangesToBranch
     git config --local user.name $CommitUsername
     git config --local user.email $CommitEmail
 
-    execute-command { & git checkout -B $BranchName }
+    # Redirect git's stderr to stdout then to Write-Host
+    # for powershell v5.1.17763.1 (Windows Server 2019) to stop throwing errors
+    execute-command { & git checkout -B $BranchName 2>&1 | Write-Host }
```
PeterGerrard: This changes behavior of this function. Git output now ends up separated most goes onto the output stream but this is directly written to the host. We should be consistent at the very least. I think you should just redirect standard error to the output stream.
```suggestion
    execute-command { & git checkout -B $BranchName 2>&1 | Write-Output }
```

PeterGerrard: This changes behavior of this function. Git output now ends up separated most goes onto the output stream but this is directly written to the host. We should be consistent at the very least. I think you should just redirect standard error to the output stream.
```suggestion
    execute-command { & git checkout -B $BranchName 2>&1 | Write-Output }
```

PeterGerrard: This changes behavior of this function. Git output now ends up separated most goes onto the output stream but this is directly written to the host. We should be consistent at the very least. I think you should just redirect standard error to the output stream.
```suggestion
    execute-command { & git checkout -B $BranchName 2>&1 | Write-Output }
```

PeterGerrard: This changes behavior of this function. Git output now ends up separated most goes onto the output stream but this is directly written to the host. We should be consistent at the very least. I think you should just redirect standard error to the output stream.
```suggestion
    execute-command { & git checkout -B $BranchName 2>&1 | Write-Output }
```

PeterGerrard: This changes behavior of this function. Git output now ends up separated most goes onto the output stream but this is directly written to the host. We should be consistent at the very least. I think you should just redirect standard error to the output stream.
```suggestion
    execute-command { & git checkout -B $BranchName 2>&1 | Write-Output }
```

nicolasvan: Actually, I don't think this changes the behavior of the function. `git checkout` seems to send its output to `stderr` which powershell will happily print on the console but won't return as part of the result of the function.

So `Write-Host` does preserve that behavior...

I tried with Write-Output first but then errors still get thrown 🤷‍♂️

PR #110, opened 2018-11-01
Only include NUnit process when running code coverage

mark-raymond:
Tests that use LocalDB result in dotCover doing code coverage of sqlservr.exe - and consequently waiting for the LocalDB instance to exit before the code coverage is finished! Which normally takes 5-10 mins. In most situations, you only actually want to cover what's run by the tests themselves, and not external processes, so this PR defaults the process filter parameters in `Invoke-NUnitForAssembly` and `Invoke-NUnit3ForAssembly` to only include the NUnit process itself.

As this is a potentially breaking change, I've bumped the version number to 1.0.

As I've bumped the version number to 1.0, I thought it made sense to include another breaking change, of fixing the typo `-DotNotImportResultsToTeamcity` -> `-DoNotImportResultsToTeamcity` (which is only used in https://github.com/red-gate/honeycomb-wpf/blob/master/.build/build.ps1).


RobClenshaw:
Seems inherently sensible!

PR #113, opened 2018-12-21
Allow overriding NUnit's process isolation setting

mark-raymond:


mark-raymond:
@PeterGerrard I've made those changes.

PR #114, opened 2019-01-03
Cmdlet to normalize AssemblyInfo.cs files

mark-raymond:
This PR introduces the `Rewrite-AssemblyInfos` cmdlet, which rewrites all AssemblyInfo.cs files in a solution in a standard format. This was originally written for SQL Clone, but has had the SQL Clone specific code removed and tests added, and it should now be useful for other products.


mark-raymond:
Looks like it's still working in SQL Clone: https://buildserver.red-gate.com/viewLog.html?buildId=9786905&buildTypeId=SQLClone_Build

PR #119, opened 2019-02-07
Support projects as well as packages.config

mark-raymond:
Main aim of this PR is to make `Update-NuspecDependenciesVersions` support 2017-style Visual Studio projects as well as packages.config files.


Public/Nuget/Update-NuspecDependenciesVersions.ps1
```diff
@@ -30,7 +34,12 @@ function Update-NuspecDependenciesVersions {
         Get-CallerPreference -Cmdlet $PSCmdlet -SessionState $ExecutionContext.SessionState -Name 'VerbosePreference'
 
         # Load the nuget packages
-        $nugetPackages = Get-NugetPackagesFromConfigFiles -PackagesConfigPaths $PackagesConfigPaths
+        $nugetPackages = if ($PackagesConfigPaths) {
+            Get-NugetPackagesFromConfigFiles -PackagesConfigPaths $PackagesConfigPaths
+        }
+        else {
+            Get-NugetPackagesFromProjects -ProjectFilePaths $ProjectFilePaths
+        }
```
RobClenshaw: Is it possible someone might want to use both packages.config files and projects in the same call to the cmdlet?

mark-raymond: Done!

PR #120, opened 2019-02-20
Fix nuget package signing

nyctef:
@PeterGerrard found a problem where the usage client nuget package was restoring incorrectly on linux (the extracted files ended up with literal backslashes in their paths instead of being put into folders, which meant the dlls couldn't be referenced correctly).

I've created this branch and [a matching branch in `RedGate.Usage.Client`](https://github.com/red-gate/RedGate.Usage.Client/compare/fix-nuget-package-signing) - IIRC the fix worked in Peter's test, but need to doublecheck that

I'm not sure if there's a good way to write automated tests for this - any ideas? Suggestions on powershell style welcomed as well :)


Public/Invoke-SigningService.ps1
```diff
@@ -218,15 +218,14 @@ function Invoke-SigningServiceForNuGetPackage {
     $TempDir = New-TempDir
     try
     {
-        # First copy the NuGet package to the temp dir, giving it a .zip extension (Expand-Archive requires a .zip extension).
```
ChrisLambrou: You've removed the comment that explains why the NuGet package had to be copied to a temp file with a `.zip` extension _(because `Expand-Archive` requires it)_, yet the code still copies the package to a temp file anyway, and the rest of the comments still refer to it as a zip file, which is no longer true.

We could/should maybe do one of the following:
- Add a comment to indicate that we copy to a temp file so that if anything goes wrong, the original file is unlikely to be corrupted. However, if anything goes wrong, any build should probably immediately abort anyway, so there's not much benefit in this.
- Stop using a temporary intermediate file and work with the original NuGet package in-situ instead.

And of course, either way, we should fix the rest of the inline comments in this cmdlet.

Public/Invoke-SigningService.ps1
```diff
@@ -259,6 +258,74 @@ function Invoke-SigningServiceForNuGetPackage {
     }
 }
 
+# based on https://stackoverflow.com/a/10072887/895407
+function Expand-OPCPackage {
```
ChrisLambrou: This cmdlet feels generically useful. At the very least, it should be moved to its own file in the `Private` folder. Maybe even become a public cmdlet, but maybe that can wait until we have an actual need for it. Same goes for the corresponding `Compress-OPCPackage` cmdlet.

Public/Invoke-SigningService.ps1
```diff
@@ -259,6 +258,74 @@ function Invoke-SigningServiceForNuGetPackage {
...
+        [string]$InputPath,
+        [Parameter(Mandatory=$true)]
+        [string]$OutputDir
+  )
+    write-verbose "Extracting package $InputPath into $OutputDir"
+
+    add-type -AssemblyName 'windowsbase'
+    $package = [io.packaging.package]::Open($InputPath);
+
+    $package.GetParts() | foreach {
```
ChrisLambrou: And regarding PowerShell code style, it'd be best to follow the style of the rest of the RedGate.Build module. It basically boils down to using the original casing for all cmdlets rather than lower case, and use upper-camel case for identifiers. Lower-case seems to be okay for keywords and aliases (e.g. `foreach`, though actually we generally tend to favour `ForEach-Object`, as far as I can see).

nyctef:
> Also, what's an OC Package?

I don't know if it's a real thing, but googling "OPC package" gives https://en.wikipedia.org/wiki/Open_Packaging_Conventions as the first result, so I picked that as the name :)

> And regarding testing, when you can't easily write automated Pester tests, I've usually tested manually on a local build. For example, in SQL Compare's `Build` function, I'm doing the following that allows the restored `RedGate.Build` package to be overridden:

Ooh, that's a neat idea - I mostly tested this by dot-sourcing the file and playing around with the functions directly, but that seems like a cool way to do more real-world testing

Public/Invoke-SigningService.ps1
nyctef: Yeah, I kept the copying logic since it seemed vaguely safer (eg we're probably less likely to get random file locking errors if we update the file atomically) but didn't think about it too hard.

I guess another difference is that we create an entirely new nupkg and replace the original instead of trying to edit one, which also feels safer - not sure if it actually makes a difference though.

I'll have a look at fixing the comments

Public/Invoke-SigningService.ps1
nyctef: Sounds like a good idea 👍

Public/Invoke-SigningService.ps1
nyctef: Ok, will clean some of that up, thanks

PR #122, opened 2019-02-27
Improvements to the ZipArchive cmdlets

ChrisLambrou:
- Improved escaping of paths when invoking external 7-zip command-line executable.
- `Expand-ZipArchive` now also uses the 7-zip command-line executable, resulting in two benefits:
  - It's substantially faster then the built-in .NET archive extraction capability.
  - It supports archive file formats other than .zip.


samblackburn:
Would be nice to see a test around this.  Happy to merge for now and let that be a separate PR, though

ChrisLambrou:
Thanks, Sam! :)

ChrisLambrou:
> Would be nice to see a test around this. Happy to merge for now and let that be a separate PR, though

Yeah, I guess the original author didn't write a test, perhaps because it trivially delegated to `Expand-Archive`. I manually checked that I hadn't introduced a regression when I made this change, but we really could do with an automated Pester check for this. It's on my magical list of to-do post-its on my monitor. I expect I'll get around to this on Friday afternoon.

PR #123, opened 2019-03-06
Added Set-TempDirPath

ChrisLambrou:
Added `Set-TempDirPath` that allows consumers to change the default location for temporary files and folders.

The default location is a `.temp` sub-folder within the RedGate.Build module. When the module is consumed internally by Redgate as a NuGet package, temp folders often look something like the following:

```
D:\BuildAgentB\work\c33c4bd13a1203ac\Engine\.build\packages\RedGate.Build\tools\.temp\d50qusse.bkc\
```

That path is already quite long. Depending on the build operation involved, there's a real risk of hitting the 260 character path limit in the Windows filesystem. This new `Set-TempDirPath` cmdlet allows you to alter the root folder in which new temporary folders are created. For example, by calling something like `Set-TempDirPath "$MyCheckoutRoot\.temp`, you can shorten the names of temporary folders to something more like the following:

```
D:\BuildAgentB\work\c33c4bd13a1203ac\.temp\d50qusse.bkc\
```


.gitignore
```diff
@@ -3,5 +3,6 @@
 Pester
 TestResults.xml
 .temp/
+.custom-temp/
```
ChrisLambrou: We use this custom folder in a new test, so best exclude it from source control.

Private/_Init.ps1
```diff
@@ -4,8 +4,6 @@
 # Create the packages folder where nuget packages used by this module will be installed.
 $PackagesDir = New-Item -Path "$PSScriptRoot\..\packages" -ItemType Directory -Force
 
-$TempFolderPath = New-Item "$PSScriptRoot\..\.temp" -ItemType Directory -Force | Select -ExpandProperty FullName
```
ChrisLambrou: This is replaced by a script-local variable in the new cmdlet.

Public/Set-TempDirPath.ps1
```diff
@@ -0,0 +1,30 @@
+$script:TempDirPath = mkdir "$PSScriptRoot\..\.temp" -Force | Select -ExpandProperty FullName
```
ChrisLambrou: This is the new script-level variable used to store the root folder for new temp folders.

Public/New-TempDir.ps1
```diff
@@ -8,7 +8,7 @@
 #>
 function New-TempDir
 {
-    $Path = "$TempFolderPath\$([System.IO.Path]::GetRandomFileName())"
+    $Path = [IO.Path]::Combine((Get-TempDirPath), [IO.Path]::GetRandomFileName())
```
ChrisLambrou: `Path.Combine` deals with any potential directory separator char issues.

Public/New-TempDir.ps1
```diff
@@ -8,7 +8,7 @@
...
     Write-Verbose "Creating temp dir: $Path"
-    return [string] (mkdir $Path)
+    return (mkdir $Path).FullName
```
ChrisLambrou: Using `FullPath` deals with the temp dir path being relative, rather than absolute.

Public/Set-TempDirPath.ps1
```diff
@@ -0,0 +1,30 @@
...
+    param(
+        [Parameter(Mandatory = $True, ValueFromPipeline = $True, Position = 0)]
+        [ValidateNotNullOrEmpty()]
+        [string] $Path
+    )
+
+    $script:TempDirPath = $Path
+}
+
+# Only the above function is public. This one is private.
```
juliahayward: Is there any point in this (as a private function)? Not sure I know Powershell well enough to have a way of TempDirPath only being accessible via these two methods rather than scripts being able to mutate it

ChrisLambrou: `Set-TempDirPath` is public, because that's the whole point. `Get-TempDirPath` is private because:

- That's how the RedGate.Build module works. Only cmdlets in the `Public` folder with a name that matches exactly with their corresponding file name are exported. See [here](https://github.com/red-gate/RedGate.Build/blob/master/RedGate.Build.psm1#L18).
- Putting `Set-TempDirPath` and `Get-TempDirPath` in the same file means that they can share the same script-level variable. I can't find any other reasonable, simple way to share state across files without the shared state being global. Unfortunately, "module-scope" doesn't seem to be a thing in PowerShell.
- There's no _current_ need for `Get-TempDirPath` to be public, so I can get away with the hack for now.

Yes it's basically slightly hacky, but hacky in a way I think we can live with for now.

juliahayward: That's fine - it looked like global state to me, it makes a lot more sense if script-scope.

PR #129, opened 2019-04-15
Change the way we execute dotcover so PS escaping is not a problem

IvoMiller:
Executes dotcover via Start-Process so that we don't have escaping problems for nunit arguments.

Arguments with spaces such as --where="cat != SomeCategory" wouldn't work and we never found a way of successfully escaping it.

Worked on this with @PeterGerrard 

This solution requires us to redirect the process output to file and then back to the console which is a bit sad.


Public/Invoke-DotCoverForExecutable.ps1
```diff
@@ -79,9 +79,15 @@ function Invoke-DotCoverForExecutable {
 
   $DotCoverPath = Get-DotCoverExePath -DotCoverVersion $DotCoverVersion
 
-  & $DotCoverPath $DotCoverArguments
+  $tempOut = New-TemporaryFile
+  $tempErr = New-TemporaryFile
```
ChrisLambrou: I'd consider using `New-TempDir` from `RedGate.Build` (or even add a new `New-TempFile` cmdlet), so that the temp files are created within the working folder, rather than polluting the current user's temp folder. This has been a problem in the past on build agents when repeated tests create many temp files/folder, and we hit maintenance issues.

Public/Invoke-DotCoverForExecutable.ps1
```diff
@@ -79,9 +79,15 @@ function Invoke-DotCoverForExecutable {
...
+  $process = Start-Process $DotCoverPath $DotCoverArguments -Wait -NoNewWindow -RedirectStandardOutput $tempOut.FullName -RedirectStandardError $tempErr.FullName -PassThru
+  $process.WaitForExit()
 
-  if ($LastExitCode -ne 0) {
-    throw "dotCover exited with code $LastExitCode."
+  Get-Content $tempOut
+  Get-Content $tempErr | Write-Warning
+
+  if ($process.ExitCode -ne 0) {
+      throw "dotCover exited with code $($process.ExitCode)."
```
ChrisLambrou: Possible regression? When the covered process writes to stderror, but has an exit code of 0. In the past, would that error output have caused an error record to be raised, and thus caused a _(desirable)_ build failure? If so, we're now emitting that error build output as a mere warning, and the build would no longer fail.

IvoMiller: I ran some tests and dotcover does fail if nunit tests fail. I have changed this to Write-Error anyway since that seems more correct.

IvoMiller:
@ChrisLambrou Any chance you could take another look?

ChrisLambrou:
Hi Ivo! Sorry, I missed the notification. I'll do a re-review, then hopefully merge this.

PR #130, opened 2019-04-15
Added -WorkerCount parameter to Invoke-NUnit3ForAssembly

ChrisLambrou:
Added -WorkerCount parameter to Invoke-NUnit3ForAssembly, to enable users to control the number of worker threads used when running parallelizable tests, as per the following documentation:

https://github.com/nunit/docs/wiki/Console-Command-Line


Private/NUnit/3/Build-NUnit3CommandLineArguments.ps1
```diff
@@ -38,5 +39,9 @@ function Build-NUnit3CommandLineArguments {
         $params += "--process=$ProcessIsolation"
     }
     
+    if ($WorkerCount -gt 2) {
```
PeterGerrard: This feels wrong, if someone specifies 1 or 2 surely that would suggest they want exactly that many, not the default value

ChrisLambrou: I dug further into the console runner source code. It seems that 0, 1 and 2 are all valid values, and actually mean different things. I'm going to mirror the source code's behaviour. Default the value to -1, and then specify it if the user has provided a value >= 0.

PeterGerrard: Sounds good

ChrisLambrou: Good to go, now, I think.

ChrisLambrou:
And the irony is, I tried this change in my tests, hoping for a build time improvement. Nada! 😢

PR #133, opened 2019-04-26
Add cmdlets for getting licenses

Greg-Smulko:
Moved from SQL Data Catalog repo + added documentation.

Initially added in https://github.com/red-gate/SqlDataCatalog/pull/1491 - see this PR for the full git history.

Tested in SQL Data Catalog by doing these changes:

* paket.dependencies:
```
nuget RedGate.Build ~> 3.5.45 prerelease
```

* licenses.tasks.ps1:
```
task PackageLicenses ImportRedgateBuildTools, {
    if (-not $LicensesBuildDir) {
        throw 'The $LicensesBuildDir is not defined.'
    }
    if (Test-Path $LicensesBuildDir) {
        Remove-Item $LicensesBuildDir -Recurse -Force
    }
    $Null = mkdir $LicensesBuildDir

    Write-Host "Licences output directory is $LicensesBuildDir"

    $slnPath = Join-Path $SolutionDir 'RedGate.SqlDataCatalog.sln'
    Invoke-GetLicensesForNuGetPackages $slnPath

    $packageJsonPath = Join-Path $SolutionDir 'WebClient' | Join-Path -ChildPath 'package.json'
    Invoke-GetLicensesForNodeModules $packageJsonPath
}
```
and then just running:
```
.\.build\_init.ps1
Build PackageLicenses
dir .\Build\3rdPartyLicenses\
```


Public/Licenses/Invoke-GetLicensesForNuGetPackages.ps1
```diff
@@ -0,0 +1,158 @@
...
+            # new nuget package format with a license expression specified
+            elseif ($_.licenseType -eq 'expression') {
+                $licenseExpression = $_.licenseValue
+                $newFileName = "$packageName@$($_.version)-license-$licenseExpression.html";
+                $destination = Join-Path $tempDirForAllLicenses $newFileName
+                Write-Verbose "Destination: $destination"
+
+                Add-Type -AssemblyName System.Web #to be able to use UrlEncode
+                # read more on https://docs.microsoft.com/en-us/nuget/reference/licenses.nuget.org
+                $licenseUrl = "https://licenses.nuget.org/" + [System.Web.HttpUtility]::UrlEncode($licenseExpression)
```
ChrisLambrou: Instead of importing `System.Web', isn't there a method on `Uri` to do this? Something like `[uri]::EscapeUriString($licenseExpression)` or `[uri]::EscapeDataString($licenseExpression)`? It'd help move closer towards PowerShell 6 support, rather than away from it.

Public/Licenses/Invoke-GetLicensesForNodeModules.ps1
```diff
@@ -0,0 +1,72 @@
...
+  Get licenses for node modules
+
+.DESCRIPTION
+  Downloads and compresses licenses for node modules specified in a provided packages.json file (including dependencies).
+  Also generates a report of all packages used, together with their licenses.
+
+.PARAMETER PackageJsonFilePath
+  A full path to a packages.json file to get licenses for.
+#>
+function Invoke-GetLicensesForNodeModules {
```
ChrisLambrou: The cmdlet name feels a bit weird. Wouldn't `Get-LicensesForNodeModules` be a bit more straightforward? Except it doesn't seem to return anything. Having reread this PR, I'm actually still not sure what the outcome is of calling it.

Greg-Smulko:
Thanks for the review!

> I'm presuming this is probably still work in progress, right?

Well, yes, and it'll be for some time most likely - see the list of TODO tasks on the [OSS Compliance board](https://github.com/orgs/red-gate/projects/23).

Public/Licenses/Invoke-GetLicensesForNodeModules.ps1
Greg-Smulko: Yes, it's a fair point.
This cmdlet does a lot of things - maybe even too many...
Basically, it writes these files to a disk:
image

It's something like `Invoke-DownloadLicensesAndGenerateReportsForNodeModules`.
In the near future, it will probably also fail a build if a license type is not allowed.
Maybe `Invoke-HandleLicensesForNodeModules`?
Thoughts?

ChrisLambrou: I'd consider changing tack. There's an ethos to PowerShell that's inherited from its Unix roots, which is that cmdlets should do just one thing, and do that one thing as well as possible. In this case, I think it would be a good approach to separate out all of those responsibilities into different cmdlets. 

Here's an example of a set of possible cmdlets. I'm not suggesting this is exactly what to aim for, but it hopefully gives an idea of how the license functionality could be delivered:

- `Get-LicenseInfo -NuGetPackageId <id> -NuGetPackageVersion <version>`: Returns the license info for a single NuGet package. The license info could be a `PSCustomObject` with a common set of properties shared by the other cmdlets I describe here.
- `Get-NuGetPackages -ProjectFilePath` or `Get-NuGetPackages -SolutionFilePath`: Returns all of the NuGet packages referenced directly or indirectly by a single project or solution. If the return type is a `PSCustomObject` with `NuGetPackageId` and `NuGetPackageVersion` properties, you can pipeline this into the previous cmdlet.
- `Get-LicenseInfo -NodeModuleId <id> -NodeModuleVersion <version>`: Does the thing that `Get-LicenseInfo -NuGetPackageId <id> -NuGetPackageVersion <version>` does, but for Node modules _(in this example, it'd actually be the same cmdlet, which has its pros and cons)_. In particular, the returned license info object should share the same properties as for the NuGet package case, so that some of the upcoming cmdlets don't need to care where the information came from, NuGet or Node.
- `Get-NodeModules -ProjectJsonFilePath`: Returns all of the modules referenced directly or indirectly by a NodeJS packages.json file. Again, if the return type is a `PSCustomObject` with `NodeModuleId` and `NodeModuleVersion` properties, you can pipeline this into the previous cmdlet.
- `Test-LicenseInfo`: This allows you to pipeline previously acquired licence info objects and perform an acceptability check on each license (i.e. is it an acceptable OSS license).

Once you have these cmdlets, the functionality behind them all is very decoupled, it's really easy to glue them together in useful ways, the consumers are free to choose exactly how to process it all, **and** you also have some really useful cmdlets that can be used for other completely different purposes.

Anyway, I'm not sure I'd do it just as I've described above. My main point is I think it would be good to decouple the whole problem into small reusable parts rather than doing everything in a single cmdlet in a very prescribed way.

Public/Licenses/Invoke-GetLicensesForNuGetPackages.ps1
```diff
@@ -127,7 +127,7 @@ function Invoke-GetLicensesForNuGetPackages {
 
                 Add-Type -AssemblyName System.Web #to be able to use UrlEncode
```
ChrisLambrou: I presume you no longer need this, now?

Greg-Smulko: Obviously, thanks. Removed.

Public/Licenses/Invoke-GetLicensesForNodeModules.ps1
Greg-Smulko: I think that what you've described makes a lot of sense. I'll give it a go.

On the other hand, I'd really like to expose a simple, ready-to-use cmdlet that just does the whole job. I assume that the simpler it would be to consume, the easier the adoption is.

ChrisLambrou: Yeah, it's a double-edged sword. An all-in-one cmdlet is easy to consume, but it takes us back to the days of "build magic", where no-one really knows what's going on under the covers and is scared to deviate from any established cargo-cult. OTOH, when the cmdlets are much simpler, it's far less obvious how to glue them together to do something useful. I'm not sure what the best way around this is, other than to provide some good examples in the cmdlet docs.

Greg-Smulko: I've extracted some bits, [and retested](https://buildserver.red-gate.com/viewLog.html?buildId=10238869&buildTypeId=CustomerDelight_SqlEstateManager_Web_Product&tab=artifacts#!tjncs,1iudsdh7pqiv6).
I'd really like to have it merged as it is, and do additional improvements in parallel with adopting it in the products. It would also allow others to contribute.
I'd like to merge this PR early next week, if you're fine with it @ChrisLambrou.

PR #134, opened 2019-05-20
Add New-ApiPortReport cmdlet

nyctef:
...based on the build script currently in the Compare repo

Currently testing on a branch in OracleTools: https://github.com/red-gate/OracleTools/pull/552


Public/New-ApiPortReport.ps1
```diff
@@ -0,0 +1,53 @@
...
+    )
+    $Assemblies | ForEach-Object {
+        $Arguments += '-f'
+        $Arguments += $_
+    }
+
+    Write-Verbose "Executing $ApiPortExe $Arguments"
+    & $ApiPortExe $Arguments | Write-Verbose
+    if ($LastExitCode -ne 0) {
+        # TODO: figure out why Execute-Command doesn't work here
```
nyctef: Not sure what the problem is here - I tried something like 

```
Execute-Command {
 & $ApiPortExe $Arguments
}
```

but that seemed to cause the `$Arguments` array to be empty when actually run

PeterGerrard: `Execute-Command` has a parameter of `Arguments` so that will be over-ridding your `$Arguments` value

ChrisLambrou:
I couple of people around mae have commented that `New-ApiPortReport` feels like we're creating an object, rather than carrying out an action. Maybe `Invoke-ApiPort` would be a better name? It's more consistent with the other cmdlets we have for running tools (like `Invoke-NUnit...`).

nyctef:
> I couple of people around mae have commented that `New-ApiPortReport` feels like we're creating an object, rather than carrying out an action.

That sounds correct to me - the thing this cmdlet does is create a new report and return it to the caller. The fact that we actually invoke a commandline under the hood could be treated like an implementation detail. 

(I'm happy to rename it if that would be understandable to more people, though)

Public/New-ApiPortReport.ps1
```diff
@@ -31,23 +31,21 @@ function New-ApiPortReport {
...
-    Write-Verbose "Executing $ApiPortExe $Arguments"
-    & $ApiPortExe $Arguments | Write-Verbose
-    if ($LastExitCode -ne 0) {
-        # TODO: figure out why Execute-Command doesn't work here
-        throw "ApiPort exited with code $LastExitCode"
-    }
+    Write-Verbose "Executing $ApiPortExe $ApiPortArguments"
+    Execute-Command {
+        & $ApiPortExe $ApiPortArguments
+    } | Write-Verbose
```
ChrisLambrou: Doesn't `Execute-Command` already do something appropriate with its output by default? Presumably it's emitted to the console, which is fine, right? Is there any need to send it to the verbose pipeline?

Public/New-ApiPortReport.ps1
nyctef: Ah, that makes sense, thanks

(I also tried running `Execute-Command -Command $ApiPortExe $Arguments` which shouldn't have the same problem, but that also didn't work - probably for some other reason in hindsight)

ChrisLambrou: What's wrong with `exec` from Invoke-Build? Do we not use that? Can't we just do whatever happens in the other `Invoke-Xxx` cmdlets?

Public/New-ApiPortReport.ps1
nyctef: `Execute-Command` puts it output on the default pipeline, which causes a problem like this if you try to publish the report:

image

Public/New-ApiPortReport.ps1
nyctef: I tried using `exec` initially, but `RedGate.Build` doesn't appear to have access to `Invoke-Build`. `Execute-Command` seems to be the thing that `RedGate.Build` uses in general.

Public/New-ApiPortReport.ps1
```diff
@@ -0,0 +1,51 @@
+<#
+.SYNOPSIS
+  Creates an ApiPort report for .NET Standard compatibility for the passed-in assemblies.
+.DESCRIPTION
+  The report is exported as a html file, which can be published to teamcity as a build report tab.
+.EXAMPLE
+  New-ApiPortReport $Assemblies | Write-TeamcityPublishArtifact
+.OUTPUTS
+  The path to the generated html file.
```
ChrisLambrou: This is a bit odd. We should either emit the report as an object, or accept the output path as a parameter. I have a build script where I planned to run ApiPort multiple times, and was expecting to write them into a build output folder of my choosing. This cmdlet makes that rather awkward.

Public/New-ApiPortReport.ps1
```diff
@@ -0,0 +1,51 @@
...
+
+    $OutputDir = New-TempDir
+    $OutputReportPath = "$OutputDir\CompatibilityReport.html"
+
+    $ApiPortDir = Install-Package "Redgate.ThirdParty.ApiPort" "2.5"
+    $ApiPortExe = "$ApiPortDir\content\net461\win7-x64\ApiPort.exe" | Resolve-Path
+
+    $ApiPortArguments = @(
+        'analyze'
+        '-t', '.NET Standard, Version=v2.0'
```
ChrisLambrou: This framework version ought to be configurable.

nyctef: Yeah, I considered adding a few more parameters here, but it seemed like it would be trivial to do as soon as someone needed it

Public/New-ApiPortReport.ps1
ChrisLambrou: Only if you assume the output should only be the output path of the report, which is a bit weird, and certainly inconsistent with the rest of our cmdlets in this module.

Public/New-ApiPortReport.ps1
nyctef: I agree an in-memory representation would be a bit nicer, but a file path seems to be the easiest thing to work with. Couldn't you do something like `New-ApiReport | Copy-Item -Destination "output/myreport.html"` ?

ChrisLambrou: Well yes, but...

My main overall concern is a lack of consistency across the cmdlets in the module, which just amkes it harder to use. This cmdlet basically behaves a lot differently to all of the other cmdlets we have for invoking an underlying tool. For better or for worse, they've set a pattern of usage, and this cmdlet goes against the grain.

https://github.com/red-gate/RedGate.Build/blob/master/Public/Invoke-DotCoverForExecutable.ps1
https://github.com/red-gate/RedGate.Build/blob/master/Public/Invoke-NUnit3ForAssembly.ps1
https://github.com/red-gate/RedGate.Build/blob/master/Public/Invoke-NUnitForAssembly.ps1
https://github.com/red-gate/RedGate.Build/blob/master/Public/Invoke-SigningService.ps1
https://github.com/red-gate/RedGate.Build/blob/master/Public/Invoke-SmartAssembly.ps1

Public/New-ApiPortReport.ps1
nyctef: The upshot is that whatever output format you choose to have the report in, putting this extra junk on the pipeline makes it harder to consume

> , which is a bit weird, and certainly inconsistent with the rest of our cmdlets in this module.

Could you explain more about this? When writing the code here I found lots of cmdlets like `Install-Package`, `New-TempDir` and `Invoke-SigningService` which appear to work like this using file paths

Public/New-ApiPortReport.ps1
nyctef: > For better or for worse, they've set a pattern of usage, and this cmdlet goes against the grain.

So I might be missing something obvious here, but I see three patterns in the example you provided:

- `Invoke-DotCoverForExecutable` and `Invoke-SmartAssembly` take input and output file names as parameters
- `Invoke-NUnitForAssembly` and `Invoke-NUnit3ForAssembly` don't let you specify the output location, but publish the output themselves
- `Invoke-SigningService` is a bit like `New-ApiPortResult` in that it returns the output file name on the pipeline, but it modifies the file in-place so it's not really relevant

nyctef:
@ChrisLambrou Thanks for the comments - some good points raised

It sounds like there might be a compromise of splitting this into both `Invoke-ApiPort` (lower-level, lets you specify the output file name) and `New-ApiPortReport` (higher-level, same API as what's here now)

Does that sound sensible to you?

Public/New-ApiPortReport.ps1
ChrisLambrou: Yes, there's definitely some inconsistency in their API. Some accept output paths, others chose the output path for the caller (which I consider to be a bug). However, they do have the following in common:

- They're convenience wrappers around existing tools or services, with a primary purpose of making their consumption easy and consistent.
- They follow a naming pattern of `Invoke-ToolOrServiceName`, to reflect what they're doing rather than what their output is. This is consistent with the kind of cmdlets you see when you invoke `Get-Command | where { $_.Name -match '^Invoke-' }`. For example, `Invoke-Pester`, `Invoke-Sqlcmd`, `Invoke-RestMethod` and `Invoke-Chocolatey`. For all of these cmdlets, the naming reflects the execution of an underlying tool or service, rather than emphasising the outcome of the execution. That seems to be an idiom in PowerShell, and I think it would be better to be consistent with it.

ChrisLambrou:
> @ChrisLambrou Thanks for the comments - some good points raised
> 
> It sounds like there might be a compromise of splitting this into both `Invoke-ApiPort` (lower-level, lets you specify the output file name) and `New-ApiPortReport` (higher-level, same API as what's here now)
> 
> Does that sound sensible to you?

No. We should just pick one. I don't _really_ mind which. I'm just offering feedback on a PR. Take it or leave it.

Public/New-ApiPortReport.ps1
nyctef: > That seems to be an idiom in PowerShell, and I think it would be better to be consistent with it.

I see what you mean, but we have almost as many counterexamples (eg `New-ZipArchive` and `Expand-ZipArchive` instead of `Invoke-SevenZip`, `Install-Package` instead of `Invoke-NuGet`, `Push-GitChangesToBranch` instead of `Invoke-GitPush` and so on). I don't think there's enough consistency in this repo to say we should definitely do it one way or the other, which is unfortunate :/

smobs:
@nyctef What are the next steps for this PR?

nyctef:
I'm torn 50/50 on the `Invoke-ApiPort` vs `New-ApiPortReport` question - some additional opinions either way would be helpful, since it's really about what's best for the consumers. Once we have that sorted out I think this can just be merged?

smobs:
I'd lean towards specificying the path, it gives the consumer more control, and makes the least assumptions about usage.
The user is going to need to add wrapper code for finding their assemblies and publishing the artifact to TC anyway, if we wanted to share a higher level API in the future then that might be the area I'd look at.

In 6 months everything will be .net core and we can delete this code anyway.

nyctef:
Cool, will do that then

nyctef:
Updated into OracleTools and it seems to work, so merging

PR #136, opened 2019-06-17
Option to require rebases for Merge-PullRequest

nyctef:
- Pull out `Invoke-GithubRestMethod` to make Github API calls a bit simpler
- Add a `-RebaseRequired` switch to `Merge-PullRequest` (off by default)

This protects against a fairly common failure state for `Merge-PullRequest`, where eg one branch changes the signature of a method, and a second branch adds a new usage of the old signature. This doesn't cause a merge conflict since the changes are in different parts of code, and each branch builds by itself, but the combined branches will break `master`.

A version of this script has been hanging around in the `data-masking` repo for a while - never got around to merging it back into `RedGate.Build` until now 😅 

TODO:
- [x] Figure out what's going on with the test failure
- [x] Test this for a little while in the `orca-experiments` repo to make sure it still works correctly


nyctef:
We've been testing this for a couple of weeks and it appears to be working, so marking this as ready for review

Public/GitHub/Merge-PullRequest.ps1
```diff
@@ -23,45 +51,104 @@ function Merge-PullRequest {
...
-    if (-not $branch) { throw "branch name needed." }
-    if (-not $revision) { throw "revision needed. Passing an empty string here would allow any revision to be merged, which can be dangerous if more commits are pushed while the build is in progress." }
+    if (-not $branch)
+    {
+        $branch = git symbolic-ref -q --short HEAD
+        if ($LASTEXITCODE -ne 0) { throw "failed to get name for ref HEAD: is this a detached checkout?" }
+    }
+    if (-not $revision)
+    {
+        $revision = git rev-parse HEAD
```
samblackburn: Might be nice to check $LASTEXITCODE here so we don't accidentally merge the wrong revision because the command failed

Public/GitHub/Merge-PullRequest.ps1
```diff
@@ -23,45 +51,104 @@ function Merge-PullRequest {
...
+    }
 
-    $base64token = [System.Convert]::ToBase64String([char[]]$GithubApiToken);
-    $headers = @{ Authorization="Basic $base64token" };
+    Write-Host "Making sure our master branch is up-to-date"
+    git fetch origin master:master
```
samblackburn: These git commands would fail on builds with server-side checkout (instead of agent-side).  Are we just using them to Write-Host some debug info?

Public/GitHub/Merge-PullRequest.ps1
```diff
@@ -23,45 +51,104 @@ function Merge-PullRequest {
...
+    
+    $masterComparison = Invoke-GithubRestMethod -GithubApiToken $GithubApiToken -Uri "https://api.github.com/repos/$owner/$repo/compare/master...$branch"
+    $behind_by = $masterComparison.behind_by
+    if ($RequireRebase -and $behind_by -gt 0)
+    {
+        Write-Host "This branch [$branch] is behind master by $behind_by commits on github"
+        Write-Host "Attempting to rebase..."
+
+        # the build might have changed files, so we need to get rid of those changes before rebasing
+        git reset --hard
```
samblackburn: Could this break any builds that don't publish their artifacts promptly?  (hopefully not a problem as long as most artifacts are untracked)

Public/GitHub/Merge-PullRequest.ps1
nyctef: IIRC teamcity doesn't guarantee that `master`+`origin/master` are up-to-date (only the branch you have checked out will be pulled) which can cause this script to do the wrong thing.

So I think they are required, but we could potentially move them inside the `if ($RequireRebase)` check to avoid breaking anyone still using server-side checkout

Public/GitHub/Merge-PullRequest.ps1
nyctef: That's a good point, yeah - I think we tend to call `Merge-PullRequest` in a separate build configuration from anything else, so it wouldn't cause a problem in practice, but this could break things. I don't think `git reset` will touch ignored files, so if the `.gitignore` rules are set up correctly it should be fine.

Not sure if there's anything we can do about this in general, though? I guess we could try using `git rebase --autostash` as an alternative, but that would probably still break artifact publishing

PeterGerrard:
@samblackburn made some changes to address issues about calling `git`.

PR #138, opened 2019-06-19
Add initial Linux support

RobClenshaw:
This PR adds support for running RedGate.Build on Linux using PowerShell Core. I've been testing on Ubuntu 18.04 and all the tests now pass.

I've added a Linux tests TeamCity build configuration and I've moved the publish step until after both the Windows and Linux tests have run.


Public/Licenses/Invoke-GetLicensesForNodeModules.ps1
```diff
@@ -32,6 +32,7 @@ function Invoke-GetLicensesForNodeModules {
     Push-Location $tempDir
     try {
         # install license checker to a temp dir
+        New-Item -Path "./cmds/node_modules" -ItemType Directory -Force
         $Null = mkdir -p ./cmds/node_modules
```
ChrisLambrou: These two lines seem to be trying to do the same thing. Did you mean to remove the `mkdir` line?

ChrisLambrou: Also, the `$Null = ` pattern, broadly similar to `| Out-Null`, is there to suppres objects from the output pipeline. You've dropped this functionality. At best, it'll put new stuff into the log files that may or may not be an issue. At worst, it could be a breaking change, depending on the cmdlet involved.

Public/Nuget/Install-Package.ps1
```diff
@@ -29,6 +29,8 @@ function Install-Package
         [switch] $Silent
     )
 
+    $IsOnLinux = $PSVersionTable.Platform -and $PSVersionTable.Platform -eq "Unix"
```
ChrisLambrou: This simplifes to just `$IsOnLinux = $PSVersionTable.Platform -eq "Unix"`. I think I've also seen this elsewhere.

ChrisLambrou: Also, it might be worth moving this to a `Test-IsLinux` cmdlet. I'm thinking a private cmdlet, but actually it could be public.

Public/Nuget/Test-NugetPackage.ps1
```diff
@@ -26,7 +26,7 @@ function Test-NugetPackage
 
     try {
         $PackagePath = Install-Package -Name $Name -Version $Version -Silent
-        return $PackagePath -ne $Null
+        return $Null -ne $PackagePath
```
ChrisLambrou: Can simplify to `return [bool] $PackagePath`... I think, but I haven't tried this sort of stuff under Linux.

Public/Nuget/Install-Package.ps1
PeterGerrard: On PowerShell 6 there is just a global variable `$IsLinux` which might be easier to use. https://docs.microsoft.com/en-us/powershell/scripting/whats-new/what-s-new-in-powershell-core-60?view=powershell-6#engine-updates

Public/Nuget/Test-NugetPackage.ps1
PeterGerrard: That works on Linux

Public/Rewrite-AssemblyInfo.ps1
```diff
@@ -104,6 +104,10 @@ function Rewrite-AssemblyInfo {
         [Parameter(Mandatory = $true)][int] $Year
     )
     
+    if ($PSVersionTable.Platform -and $PSVersionTable.Platform -eq "Unix") {
```
ChrisLambrou: Yes, we could definitely benefit from a `Test-IsLinux` cmdlet, I think.

Public/Nuget/Install-Package.ps1
ChrisLambrou: That could be useful, but this module still has to work for all of the PowerShell 5 stuff that's currently in use by just about every product.

Public/Rewrite-AssemblyInfo.ps1
PeterGerrard: This check is actually `linux` or `mac`

ChrisLambrou: Okay, maybe we should have a `Test-IsWindows` cmdlet instead, that's less ambiguous.

Public/Nuget/Install-Package.ps1
nyctef: Does this handle macos as well? Windows tends to be the odd one out, so any logic which runs for linux should probably run for mac too

Public/Rewrite-AssemblyInfo.ps1
ChrisLambrou: Or `Test-IsUnix`?

Tests/Install-Package.Tests.ps1
```diff
@@ -5,27 +5,31 @@ $FullPathToModuleRoot = Resolve-Path $PSScriptRoot\..
 Describe 'Install-Package' {
 
     Context 'Happy Path - NUnit.Runners' {
-        if(Test-Path "$FullPathToModuleRoot\packages\NUnit.Runners.2.6.4") {
-            Remove-Item "$FullPathToModuleRoot\packages\NUnit.Runners.2.6.4" -Force -Recurse
+        $modulePath = [IO.Path]::Combine($FullPathToModuleRoot,'packages','NUnit.Runners.2.6.4')
```
PeterGerrard: It's more PowerShelly to use Join-Path (from v6 it supports more than 2 arguments, but if you need to support v5 or earlier this is probably better)

Tests/New-TempDir.Tests.ps1
```diff
@@ -2,10 +2,10 @@
 
 Describe 'New-TempDir' {
 
-    $ModuleRoot = "$PSScriptRoot\.." | Resolve-Path
-
     BeforeAll {
+        $ModuleRoot = "$PSScriptRoot\.." | Resolve-Path
         Set-TempDirPath "$ModuleRoot\.temp"
+        $PathSeparator = [IO.Path]::PathSparator
```
PeterGerrard: Should be `[IO.Path]::DirectorySeparatorChar`. This is also mis-spelt so is currently empty, and even if correct would be `:` on Linux

Public/Nuget/Install-Package.ps1
PeterGerrard: Given that `$PSVersionTable.Platform` is only available on v6 we have the same problem. A nice way to solve it would be to add something like
```
if (-not (Test-Path variable:global:IsWindows)) {
    # Before PowerShell 6 this variable isn't set and we query it. But given this only isn't set on old windows PowerShell versions we can just set it to true
    $global:IsWindows = $true;
    $global:IsLinux = $false;
    $global:IsMacOs = $false;
}
```
to the module load.

PeterGerrard: And then we can use the one for the thing that is the odd one out (generally windows)

Public/Rewrite-AssemblyInfos.ps1
```diff
@@ -53,7 +53,7 @@ function Rewrite-AssemblyInfos {
     $projects = Get-ProjectsFromSolution $SolutionFile | Where-Object { $_.Project.EndsWith('.csproj') }
     $assemblyInfos = $projects | ForEach-Object { return @{
             Project      = $_.Project
-            AssemblyInfo = $_.Project -replace '(\\|/|^)[^\\]+$', '\Properties\AssemblyInfo.cs'
+            AssemblyInfo = $_.Project -replace '(\\|/|^)[^\\/]+$', '\Properties\AssemblyInfo.cs'
```
ChrisLambrou: We've changed the regex to accomodate unix directory separators, but then appending a suffix with Windows directory separator chars in it. I reckon this ought to be using `Path.Combine`, or equivalent.

Public/Test-AuthenticodeSignature.ps1
```diff
@@ -21,5 +21,15 @@ function Test-AuthenticodeSignature {
...
+    <# TODO: PowerShell Core 6.0 does not support Get-AuthenticodeSignature
+    https://docs.microsoft.com/en-us/powershell/scripting/whats-new/known-issues-ps6?view=powershell-6
+    
+    For now, just return false as we can't test the signature, but Microsoft say this should be fixed
+    in a future release.
+    #>
+    if ($PSVersionTable.Platform -and $PSVersionTable.Platform -eq 'Unix') {
+        return $false
+    }
+
```
ChrisLambrou: It would probably be better to raise an error on non-Windows platforms, rather than lie. This would give a better indication to consumers that they can't use this functionality on non-Windows platforms.

Tests/CIServers/vsts.Tests.ps1
```diff
@@ -18,10 +18,9 @@ Describe 'Write-VSTSLoggingCommand' {
     }
 
     It 'should pass -Properties properly' {
-        Write-VSTSLoggingCommand 'area.properties' 'this is a message' -Properties @{
-            prop1 = 'value1'
-            prop2 = 'value2'
-        } | Should Be '##vso[area.properties prop2=value2;prop1=value1]this is a message'
+        $properties = [ordered]@{ prop2 = 'value2';prop1 = 'value1' }
```
ChrisLambrou: What type is `[ordered]` a shortcut for?

Tests/New-TempDir.Tests.ps1
```diff
@@ -19,8 +19,8 @@ Describe 'New-TempDir' {
             $TempDir | Should Exist
         }
 
-        It 'should contain \.temp\' {
-            $TempDir | Should Match '\\\.temp\\'
+        It "should contain \.temp\" {
+            $TempDir | Should Match "$($PathSeparator).temp$PathSeparator"
```
ChrisLambrou: `Match` uses regexes, and the Windows path separator (we should really be calling it $DirectorySeparator) is a special regex character. Isn't this causing unexpected behaviour?

ChrisLambrou: Actually, I'm not certain Pester's `Match` syntax uses regexes.

Public/Rewrite-AssemblyInfo.ps1
ChrisLambrou: Peter's approach, described elsewhere, is probably better.

Tests/CIServers/vsts.Tests.ps1
RobClenshaw: Ordered dictionary.

Tests/Install-Package.Tests.ps1
RobClenshaw: Join-Path with more than 2 args doesn't work on v5 so I felt it was safest to do this.

nyctef: The nicest thing might be to go back to the original code and just replace all the backslashes with forward slashes - forward slashes appear to work everywhere in powershell and it means there's less ugly syntax to read through for `Path.Combine` or nested `Join-Path` calls or whatever

ChrisLambrou: I'm not sure it's as simple as that. If you include forward slashes in your string literals, they could still get mixed up with backslashes when you obtain any path through the PowerShell APIs. For example, `"$PSScriptRoot/subdir/somefile.ext"` might well resolve to `"C:\GitHub\SomeRepository\.build/subdir/somefile.ext"`. If we allow inconsistent paths like that to float around in product build scripts, we're highly likely to hit all sorts of awkward bugs downstream (e.g. when passing such a path to an executable that doesn't cope with inconsistent paths).

Public/Test-AuthenticodeSignature.ps1
RobClenshaw: Thinking about it, it's not so much platform specific but Powershell version specific. Powershell Core doesn't have the cmdlet.

ChrisLambrou: Okay, so a different test and error message. But definitely raise an error rather than mislead.

You can evaluate the following to see if the command exists:

```PowerShell
[bool] (Get-Command -Name 'Get-AuthenticodeSignature' -ErrorAction 'SilentlyContinue')
```

Tests/Install-Package.Tests.ps1
nyctef: That's a good point, yeah. Afaik I've never seen an executable that doesn't cope with inconsistent paths, but I guess it is theoretically possible.

RobClenshaw:
I think I've addressed all the comments now. Anything else that I need to look at?

PR #139, opened 2019-06-20
Add support for 7-Zip on non-Windows platforms

RobClenshaw:
To support the changes I've added tests for creation and extraction of 7-Zip archives that run on both Windows and Linux on TeamCity.


Private/Zip.ps1
```diff
@@ -2,7 +2,14 @@ function Get-7ZipExePath {
...
+      $7ZipFolder = Install-Package -Name '7-Zip.CommandLine' -Version '18.1.0'
 
-  "$7ZipFolder\tools\x64\7za.exe" | Resolve-Path
+      "$7ZipFolder\tools\x64\7za.exe" | Resolve-Path
+  } else {
+      $installed = & which 7z
+      if ($installed) { "7z" } else {
+          throw "Install package p7zip-full to use 7zip functionality on this operating system."
+      }
+  }
```
ChrisLambrou: Implementation looks good. The cmdlet name doesn't match the file name, though (and I realise they didn't match prior to this PR), and it got me thinking about the name. Under Windows, it makes sense, as it returns the executable file. But under Linux, it's not really returning the path. It returns a command. Maybe the name should be something like Get-7ZipCommand instead?

ChrisLambrou: Also, the return type under Windows is `PathInfo` whereas under Linux it's just `string`. Does that inconsistency matter, given that it's internal? I'm not sure, but perhaps we could cast the Windows return value to a `string` too?

PR #145, opened 2019-11-20
Give Merge-PullRequest an optional labelName param

yarhamjohn:
Allows users to pass a labelName parameter to Merge-PullRequest instead of hardcoding it to be `merge-when-green`. The Orca team want to use the label `pr: merge-when-green` instead.

Closes https://github.com/red-gate/RedgateChangeControl/issues/925


Public/GitHub/Merge-PullRequest.ps1
```diff
@@ -75,8 +79,6 @@ function Merge-PullRequest {
         } 
     }
 
-    $labelName = 'merge-when-green'
-
     $searchUri = "https://api.github.com/search/issues?q=user%3A$owner+repo%3A$repo+type%3Apr+state%3Aopen+label%3Amerge-when-green+head%3A$branch"
```
nyctef: there's a couple of lines where `merge-when-green` is still hardcoded - here and the write-host on line 86 at least

yarhamjohn: oops

nyctef: apologies for the code being misleading 😅

yarhamjohn: Is the search uri going to work if there are spaces in the label name?

nyctef: I assume `Invoke-RestMethod` would do the right thing with escaping spaces, but it's probably worth a test

nyctef: Actually, that's probably not the case - since we're putting this into a search query string instead of the label being a separate parameter, we probably need to surround the label name with quotes

yarhamjohn:
Think this should now work with spaces in label names.

Based on https://stackoverflow.com/questions/41761927/github-api-issues-by-label-with-spaces

nyctef:
I think this is ready to be merged now - we should probably just give it a test in our repo first, though

yarhamjohn:
Works in our repo so merging

PR #148, opened 2019-12-31
Fix New-TempDir when running in a parallel build (Build-Parallel.ps1)

fredjeronimo:
# Description

This PR prevents `New-TempDir.ps1` from returning prefix garbage when running in a parallel build (`Build-Parallel.ps1`). 

[Issue 149](https://github.com/red-gate/RedGate.Build/issues/149) has been created to track future work on this.

## Details

For example, this was causing NUnit execution of a test assembly to fail if **code coverage** was enabled and if running in a parallel build:

image

The rather cryptic error in the above screenshot is issued by `New-TempDir.ps1` when called to generate a new temp location to store the standard error and standard output stream redirects from JetBrains' `dotCover`:

```
ERROR: Cannot find drive. A drive with the name 'Creating temp dir' does not exist.
```

The issue is that that `New-TempDir.ps1` writes to the current host during execution. This shouldn't be a problem, but we are actually redefining the behaviour of `Write-Host` as `Write-Output` (see [_init.ps1](https://github.com/red-gate/RedGate.Build/blob/master/Private/_Init.ps1)) due to some of the [quirks of parallel builds](https://github.com/nightroman/Invoke-Build/wiki/Parallel-Builds).

This has a big impact on any Powershell function that returns values: Any output messages get added as prefixes to the array of returned objects, which means that we no longer have a single string with the temp directory path.

In other words, instead of this:

```
D:\dev\sqlprompt\.build\packages\RedGate.Build\tools\.temp\z0y0k450.qpv
```

we would have:

```
Creating temp dir: D:\dev\sqlprompt\.build\packages\RedGate.Build\tools\.temp\z0y0k450.qpv D:\dev\sqlprompt\.build\packages\RedGate.Build\tools\.temp\z0y0k450.qpv
```

Any future I/O operations would obviously fail. We now prevent this from happening.


RELEASENOTES.md
```diff
@@ -1,3 +1,7 @@
+# 4.4
+
+- Fixed issue that caused `New-TempDir.ps1` to return prefix garbage when running in a parallel build (Build-Parallel.ps1). This was causing `Invoke-DotCoverForExecutable.ps1` to fail to execute.
```
Greg-Smulko:
```suggestion
- Fixed issue that caused `New-TempDir.ps1` to return prefix garbage when running in a parallel build (`Build-Parallel.ps1`). This was causing `Invoke-DotCoverForExecutable.ps1` to fail to execute.
```

Public/New-TempDir.ps1
```diff
@@ -9,7 +9,23 @@
...
+    # cmdlets (e.g. Write-Host) and UI members operations in functions. This is further 
+    # aggravated by the fact that we redefine Write-Host as Write-Output (see _init.ps1) 
+    # and this has an impact on the return value (i.e. the output messages get added as 
+    # prefixes to the array of objects, thus leading to failures when trying to use the 
+    # temp directory) that can cause tricky bugs.
+    # NOTE: We don't want to use other streams here (e.g. Write-Verbose) as parallel builds
+    #       employ the most basic of Powershell hosts ("Default Host"), an implementation 
+    #       of PSHost offering minimal host capabilities (the main consumer being Runspace 
+    #       itself) that only supports output. 
+    # See also: https://github.com/nightroman/Invoke-Build/wiki/Parallel-Builds
```
Greg-Smulko: How about being more specific here?
```suggestion
    # See also: https://github.com/nightroman/Invoke-Build/wiki/Parallel-Builds#avoid-host-cmdlets-and-ui-members
```

Public/New-TempDir.ps1
```diff
@@ -9,7 +9,23 @@
...
+    # See also: https://github.com/PowerShell/PowerShell/blob/master/src/System.Management.Automation/engine/hostifaces/DefaultHost.cs
+    if ($Host.Name -ne "Default Host") {
+        Write-Host "Creating temp dir: $Path"
+    }
```
Greg-Smulko: This fix is a variation of the idea described in https://github.com/nightroman/Invoke-Build/wiki/Parallel-Builds#avoid-host-cmdlets-and-ui-members .

My main worry is that we're fixing it in this specific place only, without any guarantee that it won't break in another place.
How about redefining it globally for parallel builds, as described in that article, namely:
```
# Define empty Write-Host for "Default Host"
if ($Host.Name -eq "Default Host") {
    function Write-Host {}
}
```

fredjeronimo: Yes, I thought about that as well. In fact, in SQL Prompt's side, I do just that (https://github.com/red-gate/SQLPrompt/commit/257b9d17ba3cf99774169b25cee6552e91a7de75) to get around the problem until this got addressed in `RedGate.Build`.

I share your worry, but the downside is that by doing that, we completely remove any output in the library from the results in a paralllel build. As this is only really a problem for functions that return values, I ended up prefering this approach as it avoids the complete radio silence. 

Perhaps a warning in `_init.ps1` and perhaps even in the readme for the project?

fredjeronimo: @Greg-Smulko - Thanks for the review. I've added a section on the `readme.md` and a note on `_init.ps1` about the issues your raised.

ChrisLambrou: Yeah, if `Write-Host` is a problem in parallel builds, we really ought to consider removing it entirely from RedGate.Build. I don't know what the alternative is, though, since `Write-Out` has a very specific meaning in the context of any cmdlet, so it's generally not safe to write to. Maybe a better alternative would be to convert all `Write-Host` calls to `Write-Verbose`? There is `Write-Information`, but the default behaviour is `SilentlyContinue', and changing it to `Continue` might result in the same problems you're seeing with `Write-Host`.

README.md
```diff
@@ -13,3 +13,55 @@ We use [Invoke-Build](https://github.com/nightroman/Invoke-Build)
 .\build.cmd ? # Get a list of available tasks
 .\build.cmd   # Run a build with the default tasks
 ```
+
+# (For developers) Parallel builds and host cmdlets in functions that return values
+
+When running in a parallel build (see Invoke-Build's [Build-Parallel.ps1](https://github.com/nightroman/Invoke-Build/blob/master/Build-Parallel.ps1)), we [want to avoid](https://github.com/nightroman/Invoke-Build/wiki/Parallel-Builds#avoid-host-cmdlets-and-ui-members) any host cmdlets (e.g. `Write-Host`) and UI members operations in functions as this causes a myriad of problems. 
+
+This is further aggravated by the fact that we redefine `Write-Host` as `Write-Output` (see [_init.ps1](https://github.com/red-gate/RedGate.Build/blob/master/Private/_Init.ps1)) and this has a _big_ impact on all Powershell functions that return values: Any output messages get added as prefixes to the array of returned objects, which means that we no longer have the expected outcome (e.g.when we pass it as a parameter to an external tool):
```
ChrisLambrou: > we redefine `Write-Host` as `Write-Output`

Yikes! That's potentially disastrous! We really should prioritise finding a way to not do that, and get output logging to work for both local developer builds and TC builds in a way that doesn't potentially screw up any cmdlets that happen to use `Write-Host`. The best general advice would be to avoid `Write-Host` in all cases _(except when you explicitly want to solicit user input and aren't ever running headless, but that behaviour doesn't belong in any module such as this)_.

Greg-Smulko: How about at least creating an issue for this?

Public/New-TempDir.ps1
ChrisLambrou: I think this is now resolved, in the sense that whilst we don't plan to do anything further in this PR, issue #149 has been filed and should be attended to as priority.

README.md
ChrisLambrou: Done. See #149.

PR #153, opened 2020-02-07
Add a parameter to allow --force, with --force-with-lease as the default

BLWood:
This change allows us to `--force`, which was replaced with `--force-with-lease` in [this commit](https://github.com/red-gate/RedGate.Build/pull/152/commits/de9fafd9fecb0785d6d904a1ea5002e2a0f0f7dd).

This was causing problems with the way that SCA buildmonkey works. In each successive build we push to the same branch name with different (non-linear) commit hashes, which means that we need to push with `--force`.

This adds a parameter to `Push-GitChangesToBranch` that allows this, but it defaults to the safer `--force-with-lease`.


Greg-Smulko:
So, what's the consensus after all?

Linking to https://redgate.slack.com/archives/C099Y4DH9/p1581076814070500?thread_ts=1581076814.070500&cid=C099Y4DH9

BLWood:
It seemed like adding this parameter to `Push-GitChangesToBranch` was not the best solution, since the first call we make to `Push-GitChangesToBranch` is through `Update-RedgateNugetPackages`. So we would either need to make `Update-RedgateNugetPackages` force push, or change the order of our pushes.

We decided instead to do a `git fetch` every time before we push, which resolved the issue nicely. So I've closed this PR.

PR #159, opened 2020-02-25
[Cleanupcode]: bump version, modify commit message, update docs

Greg-Smulko:
[Cleanupcode]: bump version, modify commit message, update docs


Greg-Smulko:
Tested on SQL Monitor code, it seems that the change below is the only one caused by the version bump:
image

PR #160, opened 2020-03-11
Stricter revision checks in Merge-PullRequest

nyctef:
A while back we saw an apparent case where Merge-PullRequest merged the wrong commit: https://github.com/red-gate/RedgateChangeControl/issues/1156

We haven't seen this reoccur - it seems to be a very rare race condition or something - but I was just picking this issue up and saw a couple of changes we could make here.

From the build log, it appears that `$revision` was empty, although I have no idea why:
```
15:08:18  Local branch: hot-loading | Local revision: 
15:08:18  Local master: 218cbbbf5eaa334a8a92774e3fed3cd25e8f2dac | Local origin/master: 218cbbbf5eaa334a8a92774e3fed3cd25e8f2dac
```

Switching the `$revision` check to `.Equals` instead of `.StartsWith` should mean that `Merge-PullRequest` fails in a safer way if this happens again.

Also, we pass `$revision` instead of `$pullSha` to the github PR merge API, so the PR should only be merged if it still matches the current revision. Technically `$revision` and `$pullSha` should be equal at this point, but this might protect against future refactoring introducing a similar issue.


dnlegge:
LFTM. Moving from starts with to equals feels safer

nyctef:
It's potentially a breaking change (the old code seems to be designed to cope with someone passing in a short hash to the `$revision` parameter) but will merge in a day or so if nobody shouts

PR #161, opened 2020-03-18
Allow npm loglevel (used during license checker installation) to be specified as an argument

ojhall:
At time of writing, both SQL Clone and SQL Data Catalog depend indirectly on the npm package mkdirp@0.x.y. Version 0 has recently been marked as legacy, so installing it issues an npm warning. We aren't in control of the version of mkdirp our dependencies ask for, so there's little we can do except wait for them to update mkdirp. But the warning gets in the way of our builds.

This change enables that warning to be ignored by setting the loglevel to "error" or higher until the entire dependency mess sorts itself out.

This code has been tested by overwriting our local copy and testing the relevant build step.


fredjeronimo:
👍 LGTM

Public/Licenses/Invoke-GetLicensesForNodeModules.ps1
```diff
@@ -33,7 +39,8 @@ function Invoke-GetLicensesForNodeModules {
     try {
         # install license checker to a temp dir
         New-Item -Path "./cmds/node_modules" -ItemType Directory -Force | Out-Null
-        npm.cmd install --prefix ./cmds license-checker
+        $logLevelArgs = if ($LogLevel) { @('--loglevel', $LogLevel) } else { @() }
+        npm.cmd install --prefix ./cmds license-checker $logLevelArgs
```
nyctef: I thought you needed to splat the array like `@logLevelArgs` rather than just passing in `$logLevelArgs`, but it's powershell so I wouldn't be surprised if this works too :)

fredjeronimo: Oh, the joys of PowerShell. The outcome in this case seems to be the same:

image

ojhall: Interesting, I've never seen that before. Is it literally called 'array splatting'? Are there docs?

fredjeronimo: https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_splatting?view=powershell-7

fredjeronimo: I'm guessing that even if both work, splatting is probably the better way...

ojhall: I'll be on a call for the next hour but feel free to update this.

PR #162, opened 2020-05-21
Use a newer version of ApiPort

adrianbanks:
The version currently in use is an old version. The [latest version](https://github.com/microsoft/dotnet-apiport/releases/tag/2.6.0-alpha) is 2.6.0-alpha, which we already have wrapped in a different NuGet package.


Greg-Smulko:
Does not https://github.com/red-gate/nuget-thirdparty/blob/f9edba991505372b8e83a98459c0fd1916ee1484/.build/_init.ps1#L118 need to be updated too, as the name of the package changes?

adrianbanks:
@Greg-Smulko Not sure what you mean.

We were using the `Redgate.ThirdParty.ApiPort` package which looks like it was made manually and there is only one version of. I've switched to using the `RedGate.ThirdParty.Microsoft.Dotnet.ApiPort` package which is built using a script (the one you linked to) and there is a new version of (the version I want to use).

I can't see why I'd need to change the script you linked to.

PR #164, opened 2020-05-22
Add capability to profile individual tests

mark-raymond:
This PR adds capability to Invoke-NUnit3ForAssembly to run profiling, using [RedGate.TestProfiling](https://github.com/red-gate/RedGate.TestProfiling) and [dotTrace](https://www.nuget.org/packages/JetBrains.dotTrace.CommandLineTools.windows-x64).

I have tested this using a branch of [RedGate.SqlServer.ConnectionControl](https://github.com/red-gate/RedGate.SqlServer.ConnectionControl/compare/profile?expand=1) – see the [TeamCity build](https://buildserver.red-gate.com/viewLog.html?buildId=12873479&tab=buildResultsDiv&buildTypeId=SharedComponents_RedGateSqlServerConnectionDialog), which has a PerformanceTestResults.zip with the dotTrace results in.


Public/Invoke-DotTraceForExecutable.ps1
```diff
@@ -0,0 +1,70 @@
...
+
+  if ($TargetArguments) {
+    $DotTraceArguments += $TargetArguments
+  }
+
+  $DotTracePath = Get-DotTraceExePath -DotTraceVersion $DotTraceVersion
+
+  $process = Start-Process $DotTracePath $DotTraceArguments -Wait -NoNewWindow -PassThru
+  $process.WaitForExit()
+
```
PeterGerrard, line 66: Do you want to do anything with the process output?
https://github.com/red-gate/RedGate.Build/blob/master/Public/Invoke-DotCoverForExecutable.ps1#L88-L89

But not to the output stream for the first line

Public/Invoke-NUnit3ForAssembly.ps1
```diff
@@ -67,6 +73,18 @@ function Invoke-NUnit3ForAssembly {
       throw "Unexpected NUnit version '$NUnitVersion'. This function only supports Nunit v3"
   }
 
+  if ($EnableCodeCoverage -and $EnableTestProfiling) {
```
PeterGerrard, line 76: Is it better to use parameter sets to say which set of options we are in?

Public/Invoke-DotTraceForExecutable.ps1
mark-raymond: ...you make a good point, I had sort of assumed that if it wasn't redirected it would get output, but it doesn't seem to be! TBH I'm not sure why we're using Start-Process rather than calling directly, what do you think?

Public/Invoke-NUnit3ForAssembly.ps1
mark-raymond: That would be a good idea. I hadn't done so initially because this cmdlet already has a huge complicated set of parameters, but I should probably put the effort in!

Public/Invoke-DotTraceForExecutable.ps1
PeterGerrard: Because you've captured it using `Start-Process` rather than running directly the output lives inside the `$process` object.

The `Start-Process` is used I think to make sure the cli finishes, there have been issues in the past (I think) were the cli hasn't finished but powershell continues

Public/Invoke-NUnit3ForAssembly.ps1
PeterGerrard: It only makes the the API more explicit through contract than runtime issues I would be happy for it to be added as an issue, but then maybe it won't be gotten round to.

mark-raymond: OK, I've done this now (although I've left this check in at runtime).

It does mean you just get `Invoke-NUnit3ForAssembly : Parameter set cannot be resolved using the specified named parameters.` instead of the more helpful `Code coverage and test profiling are incompatible.` 😞

Public/Invoke-DotTraceForExecutable.ps1
mark-raymond: Fixed!

mark-raymond: Ah – interestingly, I've just compared with the output before I made this change, and it was getting the output when not redirecting. I think I had done that deliberately so you'd get test output as you went. Hmm.

PeterGerrard:
Looks good, we can also now change executable logging on 1 place which is nice

PR #165, opened 2020-06-25
Invoke signing service with specific file type for PowerShell formatting files

BLWood:
Exposing the changes in the signing service added [here](https://github.com/red-gate/SigningService/pull/23).

Call the signing service with a specific file type for PowerShell formatting files, since they need to be given a different file extension to be signed correctly.


nicolasvan:
Signing service changes are deployed, so let's :shipit:

PR #166, opened 2020-07-02
Fix a bug in Zip-Archive

samblackburn:
Zip-Archive crashed when given a non-canonical base path.  This PR fixes the crash, guarded by a test.


Public/New-ZipArchive.ps1
```diff
@@ -47,7 +47,8 @@ function New-ZipArchive {
 
         Resolve-Path $Files | ForEach {
             if ($BasePath) {
-                $destination = $_.Path -replace ([Regex]::Escape($BasePath)), $tempFolder
+                $ResolvedBasePath = Resolve-Path $BasePath
```
MatthewChappelow, line 50: I would resolce the base path outside of the loop

Tests/ZipArchive.Tests.ps1
```diff
@@ -17,6 +17,21 @@ Describe 'New-ZipArchive' {
...
         }
     }
+
+    Context 'A 7zip archive created from a relative path' {
+        $TempDir = New-TempDir
+        $OutputFile = Join-Path $TempDir "Output.7z"
+
+        New-Item -Path $TempDir -Name "folder" -ItemType Directory | Out-Null
+        New-Item -Path $TempDir -Name "1.txt" -ItemType File -Value "1" | Out-Null
+        New-Item -Path $TempDir -Name "2.txt" -ItemType File -Value "2" | Out-Null
```
MatthewChappelow, line 27: I think only one file is required to reproduce the issue

PR #169, opened 2020-07-30
Extend Select-ReleaseNotes with the -LatestMinor flag

Greg-Smulko:
Extend `Select-ReleaseNotes` with the `-LatestMinor` flag, which allows getting release notes for all version that have the same `major.minor` version as the latest one.

Also, loosened the regex that's responsible for parsing line with a version number, to allow parsing SQL Monitor specific format: `## Version 10.2.2 - July 30, 2020` (as opposed to `## 10.2.2 - July 30, 2020`).

Added new tests, as well as tested manually by uploading to confluence - see https://documentation.red-gate.com/display/TS/SQL+Monitor+10.2+Release+Notes+-+embedded .

Review commit-by-commit suggested.


Public/Select-ReleaseNotes.ps1
```diff
@@ -203,7 +205,7 @@ function Select-ReleaseNotes {
         throw 'No $ReleaseNotesPath or $ReleaseNotes specified'
     }
 
-    $VersionRegex = '^\s*#+\s*(?<version>[0-9]+\.[0-9]+(\.[0-9]+)?(\.[0-9]+)?)(\s*-\s*(?<date>.*))?\s*$'
+    $VersionRegex = '^\s*#+\s*[a-zA-Z\s]*(?<version>[0-9]+\.[0-9]+(\.[0-9]+)?(\.[0-9]+)?)(\s*-\s*(?<date>.*))?\s*$'
```
Greg-Smulko, line 208: I wonder if this is not too permissive, perhaps it should be `[a-zA-Z]*\s*` instead?

Public/Select-ReleaseNotes.ps1
```diff
@@ -212,10 +214,16 @@ function Select-ReleaseNotes {
     $Accumulator = @{}
     $StraplineAccumulator = $false
     $Release = $nul
+    $LatestVersion = $nul
```
ChrisLambrou, line 217: `$nul` -> `$null` ?

ChrisLambrou: Whe has `$nul` come from in this file?

Greg-Smulko: No idea, I was just following the convention in this file. I can take a look later.

Greg-Smulko: I think it just works by accident, because `$nul` is not defined, so it's, well, `$null`... ;)

I fixed it in a5bd177.

Public/Select-ReleaseNotes.ps1
Greg-Smulko: Changed in f3a007b.

PR #172, opened 2020-09-07
Invoke-SigningService now uses a temp file

ChrisLambrou:
The Redgate signing service works by receiving an unsigned file in the body of an HTTP POST request, and the corresponding signed file is then returned in the HTTP response.

In the `Invoke-SigningService` cmdlet, this boiled down to the following call:

```PowerShell
$Response = Invoke-WebRequest `
    -Uri $SigningServiceUrl `
    -InFile $FilePath `
    -OutFile $FilePath `
    -Method Post `
    -ContentType 'binary/octet-stream' `
    -Headers $Headers
```

Over the years I've seen intermittent problems here where the `Invoke-SigningService` cmdlet would fail with the following error message:

```
ERROR: The process cannot access the file '<file-path>' because it is being used by another process.
```

Because it's been intermittent, I've never been able to reliably track down what's been locking the file. However, I've finally managed to trigger the error sufficiently consistently that I believe I've identified the cause, which is that `Invoke-WebRequest` doesn't always like the same file being specified for both the `-InFile` and `-OutFile` parameter.

This PR fixes the issue by using a temporary file for the `-OutFile` parameter, Only once the web request has completed successfully do we overwrite the input file with the signed temp file.


ChrisLambrou:
How does the versioning work for this repo? The release notes file only includes 2-part version numbers. This is a non-breaking change with no new functionality. Under semantic versioning, I'd be bumping the third digit, but there isn't one. Should I update the release notes, and if so, how?

Greg-Smulko (approved):
It's really great that you fixed it! What a timing!

Regarding release notes, I think I'd go by the example of previous changes and bump it to 5.1...

PR #174, opened 2020-10-21
Ensure temp directory exists

mircea-dogaru-redgate:
When we migrated SQL Data Catalog to Powershell 7 we updated the Redgate.Build module. Because `Invoke-SigningService` had recently been changed to write the output to a temporary file we ran into the following error: `Could not find a part of the path [file path]`. We were able to track it down to the fact that the temporary folder that we had set for the Build module with `Set-TempDirPath` did not exist.

This PR ensures the directory being set with `Set-TempDirPath` actually exists. If it doesn't exist, it gets created as part of the set operation.


Greg-Smulko (approved):
LGTM!
(I'm not sure about bumping the version number and adding the release notes, but perhaps it's worth doing?)

mircea-dogaru-redgate:
Release notes updated

Greg-Smulko (approved):
👍

